PR #21, opened 2018-02-25
Update README for folks like me

ryanleesipes:
I had a hard time figuring out exactly how to build the website, so I've added documentation to help viewers like me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,9 @@ git clone https://github.com/mozilla-l10n/bedrock-l10n locale
 git clone https://github.com/thundernest/thunderbird-notes.git thunderbird_notes
 git clone https://github.com/mozilla/product-details-json
 sudo yum install npm
-sudo npm install -g less 
-```
-#### Note: The specific version of less used and known to work is 2.7.2, it can be installed via the following command:
-```
 sudo npm install -g less@2.7.2`
 ```
+#### Note: The website will not work with less 3.0 or above.
 
 ## Run Build
 There are two build scripts, one for the [start page](https://start.thunderbird.net/) and one for [www.thunderbird.net](https://www.thunderbird.net/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,10 @@ git clone https://github.com/mozilla/product-details-json
 sudo yum install npm
 sudo npm install -g less 
 ```
+#### Note: The specific version of less used and known to work is 2.7.2, it can be installed via the following command:
+```
+sudo npm install -g less@2.7.2`
```
Sancus: I think you should just replace the generic install command, and change the note to something like "Less versions above 3.0 won't work." It's confusing to have to go back and change your less version if you accidentally run the generic command first on a distro that has Less 3 by default like Ubuntu 17.

ryanleesipes: Yeah, I'll take care of that now. Were there big change in 3.0? Can we fix that so that it does work?

ryanleesipes:
@Sancus updated, let me know what you think.

Sancus:
This looks good. You should squash your adjusted commit into the original one so there's only one commit, though.

Also I'd prefer you changed the commit message to something specific, "folks like me" is pretty vague. " Better would be: "Update README with LESS 2.7.2 and python simple server instructions."

I'm not sure if you know how to squish commits, so:

`git reset --soft HEAD~2`
`git commit -m "Update README with LESS 2.7.2 and python simple server instructions."`

After that you'll need to force push:
`git push -f origin patch-1`

kewisch:
Note you can also have github do it by clicking the merge button dropdown and selecting the squash option. We can even disallow merge commits, which we should consider doing

Sancus:
Oh huh, that option was right in front of me and I missed it. Been a year or two since I've used github cooperatively I guess. Thanks!

kewisch:
Another protip, the “allow edits by maintainers” option is on by default, which would allow you to push to Ryan’s PR branch and make changes

PR #24, opened 2018-03-01
Updated the channel page to remove mention of Earlybird

ryanleesipes:
Removed mention of Earlybird, updated other language on the page.

### DIFF
--- a/website/channel/index.html
+++ b/website/channel/index.html
@@ -16,7 +16,7 @@
  {% block content %}
    <header id="main-feature">
      <h1>{{ _('Welcome to the <strong>Future of Thunderbird</strong>') }}</h1>
-     <h2>{{ _('Download Thunderbird Beta to try the newest features in the next version of Thunderbird! By giving us your feedback, you’ll help make email better for millions of your fellow users.') }}</h2>
+     <h2>{{ _('Download Thunderbird Beta to try the newest features! By giving us your feedback, you’ll help make email better for millions of your fellow users.') }}</h2>
    </header>
  
    <main role="main">
--- a/website/channel/index.html
+++ b/website/channel/index.html
@@ -1,40 +1,41 @@
 {# This Source Code Form is subject to the terms of the Mozilla Public
- # License, v. 2.0. If a copy of the MPL was not distributed with this
- # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
-
-{% extends "_includes/base-resp.html" %}
-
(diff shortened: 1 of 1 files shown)


benbucksch:
Hey Ryan, did you change the line endings of the file? Because the diff shows the entire file, making it impossible to see what changed. It's good practice to make such mechanical changes - if they are indeed necessary - in a separate merge.

ryanleesipes:
@benbucksch I must've done that erroneously. I copied and pasted my changes into the file from another editor. Is there anything I can do to fix that now?

benbucksch:
Ryan, ah, yes.
Fix: Depends on which OS you are in. Most decent developer-oriented text editors have a way to manually change line endings. On Windows, notepad++. Also check your editors settings, to preserve the line endings of the file.
I would suggest to copy the new file to another location, undo your changes using git revert, then open both files (old and new one) using a good text editor, and copy over only the lines that you changed. A good text editor will then keep the line endings of the file you edit and not change them.

benbucksch:
Alternatively, you can simply open the file in a decent text editor, and change the line endings to "Unix", and save. Then, run "git diff 54477b65 website/channel/index.html" to verify that only those lines changed that you wanted to change.

Sancus:
The reason this hasn't been changed is because we don't have localization support yet, and changing this in the template will break every single locale because there is no translation for that phrase.

ryanleesipes:
Do we have to get contributors to translate this? Care to walk me through what that process looks like?

wsmwk:
> The reason this hasn't been changed is because we don't have localization support yet, and changing this in the template will break every single locale because there is no translation for that phrase.

Since I proposed these changes, I need to say I hadn't thought about the l10n.  These are not important changes, so I'd just  as soon wait til we have more changes for this page (link to beta community as one example), and until then focus on other more important site changes.

ryanleesipes:
Okay, but that doesn't really solve the issue of outdated pages. What needs to happen in order for this to be fixed? Where does someone contribute the updated phrases for a language?

Sancus:
Contributors do translate stuff, yes. When I said we don't have l10n support I meant we have no mechanism for them to translate anything yet, since we were relying on mozilla.org's. I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1442968 and added https://github.com/thundernest/thunderbird.net-l10n to get started on fixing that.

ryanleesipes:
We can close this pull request and keep #24 open.

PR #28, opened 2018-03-17
I need feedback on what I've done for a Get Involved page.

jerrac:
Here is a very rough outline of a Get Involved page. I've stubbed in some content, and I've added some sections without any content.

I'm not sure if I'm using the `{{ _() }}` syntax correctly. Specifically where I have inline links. Can I put the links inside the jinja syntax?

Do we actually need every single section I've added?

PR #30, opened 2018-04-03
Updated Beta Start Page

ryanleesipes:
Updated Beta start page with different wording, incorporated fontawesome into site. Can grab the fontawesome files and incorporate them locally if this is better. The new version grabs attention, and could help users find what they are looking for faster.

Feedback appreciated.


ryanleesipes:
Screenshot:
beta-start-page

ryanleesipes:
The "Get Involved" link is temporary, would like to point to #28 when it is completed.

clokep:
For comparison, here's the current (60.0b1) beta one:
screen shot 2018-04-03 at 4 24 44 pm

Sancus:
I'm like 900% sure we can just use those icons without loading 700kb of JS. Looking at their site, Font Awesome has a more complex CSS method that lets you pick and choose, it appears. If you just want to keep using that for experimentation purposes and let me handle the technical details, that's fine, I can do it after I am available again.

clokep:
@ryanleesipes I think the copy changes are nice. I like that they're a bit more excited sounding.

Additional spacing between the icons and the titles should be added.

I find the background gradient to seem a little "old", but it is better than a solid color.

wsmwk:
Initial thoughts and suggested wording changes

- c/get to experience the new Thunderbird before/get to experience improvements and fixes before/  (more explicit)
- second sentence seems long - consider splitting into two
- c/the best way to ensure/the best way for you to ensure/
- re: "Get Involved" - when most users see these words or "help out", (I think) their minds go immediately to programming/coding or "I am not good enough".  We must heavily counteract this tendency, because we need so much more than coding, and we don't need "experts".  We need l10n, we need people helping people (in SUMO, in bugzilla, etc).  That was the attempted main thrust of the old wording - I think the new wording makes a similar attempt.  But I'm wondering if there is a way to make it even more clear and even more enticing?
- Get Support - In a closing sentence can we add a link to https://discourse.mozilla.org/c/thunderbird/beta about meeting and working with other beta users, blah, blah?
- Report a bug - I want ensure beta users specify a version or identify as "beta".  The link from https://developer.mozilla.org/en-US/docs/Mozilla/QA/Bug_writing_guidelines points to https://developer.mozilla.org/en-US/docs/Mozilla/QA/Bug_writing_guidelines which does allow for version  but I wonder if that is enough.  Is https://developer.mozilla.org/en-US/docs/Mozilla/QA/Bug_writing_guidelines (which mentions "Firefox" everywhere) too overwhelming for some users, and does it provide enough context for Thunderbird (terminology), etc? 
- Submit Crash report - (the current wording is unchanged) please work in a statement of "submit every crash, including the first" crash.  One of the sentences can probably be combined.  I can provide new wording if you prefer.  Do you still need an icon for this?

wsmwk:
> But I'm wondering if there is a way to make it even more clear and even more enticing?

My initial thought about making this distinction is split into two bullets/ideas - one coding, one non-coding. In that way they have equal value.  Then point them to the same "get involved" link, or one for coding and one for non-coding.

elioqoshi:
I like the new direction! FontAwesome 5 is indeed great, but it should be kept lightweight indeed. I suppose you could incorporate SVG's of those icons and load these instead of the full FontAwesome CSS.

Some changes I'd suggest

- Headings should be bigger. The headings in the current Beta view look more suitable as they have more typographic hierarchy.

-Icons should be bigger. This goes along with the bigger headings part. To allow people to visually scan the page quickly, the icons could be larger to accompany the headings.

-Icon colors are too saturated and we might want to tweak those. Hearts and Donations are usually associated with Red and not Orange as well. A good start would be Firefox Photon colors: https://design.firefox.com/photon/visuals/color.html

I could do a mockup of this after a few days (long weekend here due to Orthodox Eastern)

elioqoshi:
Regarding the background: It would be indeed nice to update it since it's the classic Mozilla background from many years ago. We even could explore a dark variation for Nightly. I can mock these up too.

jerrac:
Here are a few thoughts.

Can you find an icon for next to "Submit crash reports"? Also, since it's not a link, it doesn't stand out at all. It feels lonesome.

A bit more space between the icons and the header text would be nice as well.

What does the page look like on a smaller screen? Do the sections move to one column when it gets skinny?

Having three sections on one side and two on the other feels unbalanced.

ryanleesipes:
The sections do move around (depending on screen-size), but the right side moves to the top. I need to look at the CSS to understand why it sorts itself out this way, but it does scale semi-properly. @jerrac as per your comments on "Submit Crash Reports", I think it would be good to include an icon. I wish I could do something with it beyond just have it present information.

elioqoshi:
@ryanleesipes would a mockup for this still be helpful?

wsmwk:
So we have URL to get involved and the beta forum.  

We should probably also need give users more info about beta. Like a couple support KBs with details on a) beta in general (FAQ?) b) beta 60, c) addons.  Perhaps we can get away with just pointing to one KB page, and that page would direct users to other KB pages as needed?

ryanleesipes:
@wsmwk I incorporated much of your feedback except the most recent bit about giving more info on the beta. I could use some input one how that would look and where it would live on the page. Also "Get Involved" coding and non-coding. I could use some coaching on how that might be worded and presented as well.

@elioqoshi I tried bigger icons, I've increased them slightly (any larger looks messed up). As I said in the comment above, any bigger changes to the start page should be done in a different Pull Request. This one is more iterative.

ryanleesipes:
@Sancus can you help me incorporate the CSS locally, I'd like to get this setup up properly. I've downloaded the relevant files from here: https://fontawesome.com/get-started/svg-with-js - but I'm not sure where they should live in the project tree.

Sancus:
Less files go in assets/less. If you have pure CSS those should go in media/css(and make a new 'start' folder).
The image files should go in media/img/thunderbird/start -- the existing icons are all together in a sprite for fast loading and that would be ideal in this case also but not sure if fontawesome supports it... if not it's fine, since this is the beta page and we can optimize that manually later if we put them on the release start page.

ryanleesipes:
Thank you @Sancus - that is what I needed.

ryanleesipes:
@wsmwk new screenshot below:
thunderbird-new-beta-with-big-icons

ryanleesipes:
@Sancus let me know how this local implementation works. (See most recent commit).

wsmwk:
Looking great. I  very much like the bigger icons - it really sets each item apart.  Mostly tiny suggestions, most of which make things shorter. Use or not, modify as needed.
- Donate section, linkfy the "donate" word in the paragraph (I don't think it will be overkill)
- Get Involved  c/help make it better/help other users, and help make it better/  c/and code to it./code, or time helping fellow users./
- Get Support  c/You're ... edges./Betas often have rough edges./    c/We have a beta category on our forum where/Visit the <beta forum> where/    then you can linkify the two words beta forum which will stand out better
- Submit Crash Reports.  I just realized the crash section is partly based on an older crash reporter. So I advocate a total rewrite as follows.   "Please report every crash. The Mozilla Crash Reporter gives valuable data on which crashes are the most serious.  And if you are comfortable doing so, please also add a public comment, and supply your email address (which is non-public) to allow us to contact you if we need to."

wsmwk:
"Mozilla Crash Reporter" can be linked to https://support.mozilla.org/en-US/kb/mozilla-crash-reporter-tb

ryanleesipes:
Okay @wsmwk - updated with the changes you proposed (or as I understood them, screenshot below).

beta-update-wsmwk

wsmwk:
Skimmed the screen shot - looks good.

ryanleesipes:
@Sancus if my implementation of the library needed for the icons looks good, I would like to request that this be merged.

Sancus:
Just moving the JS file locally doesn't fix my problem with the fact that the JS file is used at all. This should be a CSS and icons change. If that's difficult, I can look at doing it in a couple of days.

ryanleesipes:
@Sancus I've implemented the css locally, without the use of any Javascript libraries.

@wsmwk I'm looking into this. But I don't see why we couldn't roll this out if Sancus is cool with the change described above. I'll begin experimenting with different background colors for the start pages and get @elioqoshi's feedback (as he probably has a better eye for this than I do).

ryanleesipes:
Actually, hold off on rolling this in. I would like to point at the Get Involved page on #28 vs where I'm currently pointed (the old contributor link on the wiki).

ryanleesipes:
As per the most recent commit, the link has been updated. @Sancus please review.

Sancus:
Yes, looks fine with a cursory examination, will review and likely merge tomorrow.

Sancus:
This increases the size of the beta start page by ~100KB(from 144KB to 251KB). We could actually shrink it by ONLY using the fontawesome fonts and not using both open sans and the fontawesome fonts, so maybe something to consider in future patches. As this is the beta start page I'm not going to be super picky about file sizes :D

Also the .svg files are only used for iOS/Safari I believe, which would never be used to load this page, but they aren't loaded at all by Firefox/Thunderbird so it doesn't really matter.

The new icons look really nice and this is an improvement, nice work! This page does not require localization so I will send it directly to production and it should start appearing after caches expire in ~8 hours.

PR #38, opened 2018-05-18
Updated Daily start page with new icons.

ryanleesipes:
Also slightly different wording on the page.

### DIFF
--- a/start-page/daily/index.html
+++ b/start-page/daily/index.html
@@ -10,31 +10,30 @@
 {% block main %}
       <header>
         <h1>{{ self.title() }}</h1>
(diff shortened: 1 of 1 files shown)


wsmwk:
But not live?  (I'm not seeing a change)

Sancus:
@ryanleesipes Shouldn't the Get Involved section link to the new get involved page rather than the wiki?

Sancus:
Other than that, looks good, r+. Let me know if you want to change that URL or if not I'll merge it as-is.

Sancus:
https://start.thunderbird.net/en-US/daily/ This is now live.

PR #64, opened 2018-08-09
Update AMO -> ATB URLs

Croydon:


Sancus:
Thanks. I'm probably going to centralize these links in the python function, but I did forget to change change 'em.

PR #68, opened 2018-08-16
Add "email" to TB description

ryanleesipes:
Properly this time


Sancus:
I was waiting on merging this because it will mess up l10n, and I wanted to get the other website patches in, but it's fine now.

Sancus:
This change is going to need to sit on stage for at least a week or two to allow localizers to update that string btw, since it's a critical string, I don't want any regressions to untranslated versions.

Sancus:
This is live now: https://start.thunderbird.net/en-US/release/

PR #71, opened 2018-10-18
[Bug 1499831] - Add new one-time state holiday into Slovak holiday calendar

chiak597:
## Description
Add new one-time state holiday into Slovak holiday calendar

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1499831

## Testing
The ICS file is directly used as a Lightning calendar, so it is a valid ICS calendar file.


Sancus:
Thanks.

PR #75, opened 2018-10-28
Update index.lang

livingmine:
Add a space between "untuk" and "memudahkan".


Sancus:
l10n for this website is done in Pontoon, those files are vestigial and no longer used.

https://pontoon.mozilla.org/projects/thunderbirdnet/

PR #80, opened 2018-11-27
Change Twitter link

atouchet:
A link to the Thunderbird Twitter page is more convenient to use for users who don't have their own Twitter accounts or don't stay signed in.

### DIFF
--- a/website/_includes/site-footer.html
+++ b/website/_includes/site-footer.html
@@ -27,7 +27,7 @@
           </div>
           <div class="col col-3">
             <ul class="links-social">
-              <li><a class="twitter-button" title="Follow @mozthunderbird on Twitter" href="https://twitter.com/intent/follow?screen_name=mozthunderbird" data-link-type="footer" data-link-name="Twitter (Thunderbird)"><i></i><div>@mozthunderbird</div></a></li>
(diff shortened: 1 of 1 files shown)


Sancus:
I don't really agree with this. Users who don't have their own twitter accounts or who don't stay signed in are less likely to click on this, and frankly they're not important anyway because they're not going to follow, which is the main point of directing people to Twitter, so that they follow.

atouchet:
As someone who checks Twitter but doesn't use an account there I find the user experience of these follow links to be pretty poor. However if the intention of this link is mostly just to get new followers then I guess it makes sense. Feel free to close this if you prefer the existing behaviour then.

mkmelin:
I'm not a big Twitter user, but with the current link it just shows me two tweets atm. I wouldn't follow anyone without skimming through the last 10 tweets to see if they write something of interest.

Alecaddd:
I think we should merge this PR as we should change the scope of the Twitter button.
We shouldn't have a button only for the sake of getting more followers, but rather directing users to our Twitter account, which is mostly used to follow the health and activity of a company or product.

It feels more natural to have this workflow: 
`Check Twitter > Read the latest tweets and activity > Decide if to follow or not`
rather thank
`Check Twitter > Decide if to follow or not > Extra click to check the actual page and activity`

What do you think?

Sancus:
Ugh I forgot about this. I changed my mind about it a while ago. That link was supposed to actually open a little popup window, but it doesn't work without Twitter's JS include, and as a matter of security/IP privacy, I don't include external files on our sites except for Matomo.

Issue #86, opened 2019-01-03
Publicize Thunderbird Blog


wsmwk:
https://blog.mozilla.org/thunderbird/ is missing from our current public facing web pages.  (I only found it at the "old" https://wiki.mozilla.org/Thunderbird/CommunicationChannels )

Sancus:
It's not totally missing, it's in the bottom right of every release notes page: https://www.thunderbird.net/en-US/thunderbird/60.4.0/releasenotes/

So people do get kinda exposed to it when they look at those. I agree we should have a link in the main menu, how about just "News" or "Blog" in the Thunderbird menu?

kewisch:
What I wanted for the calendar site is to show the top three blog headlines and a show more link. Maybe this makes sense for Thunderbird too

ryanleesipes:
I think we should let contact us be accessible through the footer, and should give its spot to blog up at the top.

Sancus:
Agreed on the top menu.

I like the idea of having some blog headlines and stuff but I think that should be saved for the redesigned website, as I don't see an easy way to put that on the front page now with its current layout, it really doesn't fit to the left or right of the button, and putting it underneath with no other content is lame. When we have more stuff on the front page below the button, we can add a section for the blog headlines there if we want.

ryanleesipes:
Can we close this issue for now and ensure that the question you raised @Sancus is something we bear in mind for redesign?

Sancus:
Yes, I was also meaning to close this for the same reason. I've added references to this issue in the appropriate places related to the redesign, so this can be closed.

PR #95, opened 2019-03-19
Add Blog to nav bar

ryanleesipes:
As per #86 - I thought having the blog in the nav bar would be better than contact us, which can be found in the footer and frankly it would seem as though most people we want to contact us would know to look there vs the randos who have hit us up via press@thunderbird.net already asking how to fix their Thunderbird.


Sancus:
Yeah I agree with this, I was already thinking it wasn't good to have Contact Us in the header. I was also considering removing the Donate button from the footer, as it seems redundant.

PR #97, opened 2019-03-21
Fix flexbox menu for all pages

### LINKED ISSUE #96: release page layout broken by flexbox change
I should have checked every single page before merging the flexbox PR, but most were fine... except that it seems to have broken the releases page. It's only been pushed to stage so not critically urgent.

Alecaddd:
Fixes #96


Sancus:
Checked this and looks good, thanks for looking at RTL languages too.

Sancus:
This is live now, thanks again for fixes.

PR #116, opened 2019-08-20
 Rollover effects and Footer UI

Alecaddd:
* Add rollover effects to Hero area for links and buttons
* Implement basic Footer UI

### DIFF
--- a/assets/less/base/fonts.less
+++ b/assets/less/base/fonts.less
@@ -4,7 +4,7 @@
 
 body {
     font-family: @base-font;
-    &:extend(.font-base, .tracking-wide);
+    &:extend(.font-base, .tracking-wide, .text-blue-dark);
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
 }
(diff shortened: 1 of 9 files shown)


Alecaddd:
What the hell.
I pulled from redesign and squashed my latest changes before creating this PR.
Why the old commits are repeated again?
@Sancus any idea?

Alecaddd:
Maybe I'll destroy my repo and fork this again now that all the most recent changes are in the `redesign` branch.

Sancus:
Pulling only works if there's no divergence between branches. What "pull" actually means is fetch, then merge. But merging is only a good idea if there's between no diverging commits made.

When I squash a PR, that creates a divergence, because it's destroying all your old commits and making a new one that contains all their changes. So when you pulled 'redesign', all that happened was it merged the most recent changes into your 'redesign' branch. But that's useless -- you need to destroy and recreate your redesign branch.

The easiest way to do this is:
`git reset --hard upstream/redesign`

Then to add your recent changes back in:
`git cherry-pick 3a9ba794c230e92b94a3d418df593db5cf902419'

Then force push that branch to your repository:
`git push --force origin redesign`

You will need to do a hard reset and force push any time a PR is squashed to avoid creating a mess in your local branch. Alternatively, you can make a new branch for each set of changes, PR that, and then just delete that local branch afterwards.

Sancus:
The important thing to remember with git is that any time your branch and a remote branch aren't synced up, the safe way to force your local branch to be the same is to do `git reset --hard <remotename>/<branchname>`. 

Pulling isn't safe, it only works if the commit history is the exact same in both repositories.

Alecaddd:
Hell yeah, this is good to go now

PR #123, opened 2019-08-29
Add Remaining Pages

Alecaddd:
This is a Draft PR to keep track of the remaining pages.
Please, don't merge this until I change it from Draft to "Ready to be Reviewed".

- [x] All Releases
- [x] Systems & Languages
- [x] Release Notes
- [x] Organizations
- [x] Get Involved
- [x] Email Providers
- [x] 404


Alecaddd:
@Sancus @ryanleesipes any chance of reviewing this today? I need this to land in order to complete the responsive implementation.

website/404.html
```diff
@@ -6,33 +6,28 @@
...
-
-{% block body_id %}thunderbird-404{% endblock %}
-
-{% block site_header_logo %}
-  <h2>{{ high_res_img('thunderbird/landing/wordmark.png', {'alt': _('Thunderbird'), 'width': '329', 'height': '77'}) }}</h2>
-{% endblock %}
-
-{% block page_css %}
-  <link href="/media/css/thunderbird-landing.css" rel="stylesheet" type="text/css" />
+{% block page_desc %}{{ _('The page you are looking for doesn’t exist.<br/>If you think this is a bug, you can <a href="%(bug_report_url)s" class="inline-link" target="_blank">report an issue here</a>.')|format(bug_report_url=url('thunderbird.site.bug-report')) }}{% endblock %}
```
Sancus: I really dislike the idea of having classes on <a> tags that are in translated strings.... ideally these shouldn't change unless we are changing the string. That's why the page url uses a variable like that.

Alecaddd: It makes sense.
I'll go around, remove those classes, and manage everything with a container class.

Alecaddd: It should be good now.

ryanleesipes:
This is good, with one exception. I would rather see "Get Involved" up at the top now versus the 
"For Organizations". Eventually, I want "For Organizations" to be up at the top because I have plans for much more content for that page. And maybe we could do both, but I want to definitely draw attention to how folks can contribute to Thunderbird. Open to other ideas than just sticking it at the top.

Alecaddd:
@ryanleesipes sounds good, I'll change that link.

Alecaddd:
@ryanleesipes updated the link in the main nav.
I also updated the links in the footer to list "For Organization"

Sancus:
I think this is OK to merge for now.

One thing I think would be nice is if we can figure out a Javascript solution to the desire for `target="_blank"`. It's blowing up a lot of translations so far and I'm also concerned we'll just forget to change it when links change, since for a lot of links, you can change them in settings without touching the template. It seems like it's probably possible to use JS to make all links that dont contain thunderbird.net open in a tab.

PR #135, opened 2019-11-01
Create README.zh-CN.md

szxcool:
A chinese one!


Sancus:
Although I appreciate the work here, I don't think localizing our docs like this is a good idea. It's going to change frequently and possibly move elsewhere, plus cluttering the tree with a zillion READMEs in different languages doesn't feel particularly good either.

PR #140, opened 2019-11-15
Remove target attribute from translatable strings

Alecaddd:
I was going to add a quick JS snippet to dynamically add the `target="_blank"` attribute on page load for all the links that point externally, but I've no idea why this seems to happen already.
I cleared the cache and opened an incognito mode, and links that don't have a `target="_blank"` seem to get it automatically.

So we already have this sort of method in place somewhere?

### DIFF
--- a/website/404.html
+++ b/website/404.html
@@ -6,7 +6,7 @@
 
 {% block page_title_prefix %}{{_('Thunderbird')}} — {% endblock %}
 {% block page_title %}{{ _('There’s nothing here!') }}{% endblock %}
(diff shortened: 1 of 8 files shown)


Sancus:
Closing this in favor of #141. I couldn't reproduce the target blanks being added(the first two I looked at, on the 404 page and release notes page, did not have target=blank after your patch).

However, you also didn't remove most of the lines that had target=blank, so I'm confused? Did you forget a commit in this PR or something?

Anyways, I redid it in #141 and added the JS, which is definitely needed, and seems to work as expected.

PR #141, opened 2019-11-18
Remove target="_blank" from templates, and add Javascript to compensate.

Sancus:
`var bouncerURL = 'download.mozilla.org'` should probably be automatically set by the Python, so that this isn't duplicated, but it needs to be there because adding target="_blank" to the download URLs makes them janky in Firefox at least, it opens a new tab and then closes it immediately.

If this patch is good I'll add a follow-up commit for that tomorrow.


Alecaddd:
Looks good, thanks for taking care of this.

PR #152, opened 2019-11-30
Fix missing group headers in release notes for beta page

### LINKED ISSUE #151: Redesigned Release notes page doesn't support group headers for Beta
https://www-new.thunderbird.net/en-US/thunderbird/71.0beta/releasenotes/

The release notes format supports multiple groups of notes with a heading, which is mostly used for consecutive beta versions as on the live page: https://www.thunderbird.net/en-US/thunderbird/71.0beta/releasenotes/

Unfortunately, it seems like we forgot about that on the redesigned template so it looks completely broken right now. It needs support for those group headings.

Alecaddd:
This should fix the release notes page.

Fixes #151


Alecaddd:
Does this look good @Sancus?

Sancus:
Yep just checked this out locally, works great, nice job. Thanks!

PR #165, opened 2020-01-27
Add Careers Page

Alecaddd:
This PR implements the Carrers page as requested by Ryan and Magnus.


Sancus:
I feel like this would be better with a yaml back-end for the actual roles and descriptive text, so it's easier to modify, but this is fine for a first iteration. Probably some thought needs to go into the proper way to format that anyways.

@ryanleesipes Is this intended to go live ASAP or do you want it to coincide with a blog post or something?

Alecaddd:
I agree regarding the yaml file.
The markup I used for the various roles is identical and repeated, so building something like this similar to how we handle release notes it's totally doable.

website/careers/index.html
```diff
@@ -0,0 +1,320 @@
...
+          <small class="font-normal font-md tracking-normal normal-case">
+            {{ _('Full Time / Remote') }}
+          </small>
+        </aside>
+      </h3>
+      <p class="font-lg tracking-wider leading-loose mb-5 p-links-blue">
+        {{ _('Thunderbird is looking for a multi-skilled self-starter to work partly on release engineering, and partly on operating the cloud based web infrastructure around Thunderbird. You will be working in close cooperation with our release engineer, infra engineer and other staff, as well as with community members.') }}
+      </p>
+      <p class="font-lg tracking-wider leading-loose mb-6 p-links-blue">
+        {{ _('This is a <strong>remote</strong>, <strong>full time</strong> position. You are expected to have excellent written communication skills and coordinate your work over email, video conferencing, IRC, and Bugzilla.<br/>
```
Sancus: Should this be in its own <p> tag? The <br> does not actually produce a separate paragraph, it looks like this:

image

website/careers/index.html
```diff
@@ -0,0 +1,320 @@
...
+          {{ _('Help out with add-on reviews.') }}
+        </li>
+        <li>
+          {{ _('Engage with add-on developers to answer their questions.') }}
+        </li>
+        <li>
+          {{ _('Helping to integrate the best parts of popular add-ons directly into core.') }}
+        </li>
+        <li>
+          {{ _('.Helping the Thunderbird team understand the needs of add-on authors.') }}
```
Sancus: Redundant '.' at beginning of sentence

website/careers/index.html
```diff
@@ -0,0 +1,320 @@
...
+      <a href="#apply" class="btn-join self-start">{{ _('How to Apply')}}</a>
+    </aside>
+  </section>
+
+  <section class="pt-10 pb-10 flex flex-col max-w-6xl mx-auto justify-between pl-8 pr-8">
+    <aside>
+      <h3 class="header-section">
+        <span>{{ svg('bookmark') }}</span>
+        <aside class="flex flex-col">
+          {{ _('Add-ons Coordinator')}}
```
Sancus: Ryan and I agree a better name for this would be "Add-on Support Engineer" so lets change that.

website/careers/index.html
Alecaddd: That's how it was in the DOC Ryan sent me. I simply respected how it was written. I don't have any preference for this matter

ryanleesipes: Let's make it look nice, the formatting in the doc is likely just not right.

website/careers/index.html
ryanleesipes: Kill it '.'

website/careers/index.html
ryanleesipes: I think this is a reasonable update. Let's change it.

ryanleesipes:
> I feel like this would be better with a yaml back-end for the actual roles and descriptive text, so it's easier to modify, but this is fine for a first iteration. Probably some thought needs to go into the proper way to format that anyways.
> 
> @ryanleesipes Is this intended to go live ASAP or do you want it to coincide with a blog post or something?

This is a good idea. We should take this live as soon as it is in reasonable shape to do so., we can make changes if the conversation we are having on the list changes things.

PR #166, opened 2020-01-29
Update mzla references

Alecaddd:
Updates to the footer and About Us page as requested by Ryan.


ryanleesipes:
Thank you for moving so quickly on this.

ryanleesipes:
Any reason why I see this change on the main page footer but not the other pages?

wsmwk:
I see it in the footer for https://www.thunderbird.net/en-US/get-involved/ but not for https://www.thunderbird.net/en-US/about/

Sancus:
> Any reason why I see this change on the main page footer but not the other pages?

I cleared the main page cache only after I pushed that to prod, not the global cache. Clearing the global cache puts a lot of load on the servers, as it affects ALL subdomains of thunderbird.net, and more granular wildcards are locked behind the Enterprise tier of Cloudflare.

By now the caches for all pages should have expired so everything should be up to date.

PR #167, opened 2020-01-29
Fix style of form-select on non-Mozilla browsers

KexyBiscuit:
https://caniuse.com/#feat=css-appearance shows all other browsers support `-webkit-` prefix other than `-moz-`, or support nothing at all.

On WebKit browsers, there're 2 arrow on the right of `<select>`, one is from user agent (which is not hidden).


Sancus:
Yep, that does appear to be an issue. Thanks for the fix.

PR #172, opened 2020-02-11
Subheader links

### LINKED ISSUE #169: should be possible to link to sub-sections easily
Many of the pages basically lack the ability to link to subsections. This would be useful for at least linking to the beta program, or to positions on the careers page. 

A common way for this is to have the an anchoring icon appear near the relevant header on hovering. Then with the id linked, the section may want to be highlighted with css :target

Alecaddd:
Implement an easy way to add anchor links when we need them, brutally copied from the GitHub markup.
I only did a couple in the home page and all the most important in the Careers page, which is where these are most useful.

If this is a good solution, it can be implement anywhere without problems.

### Screenshot
Screen record from 2020-02-11 10 49 08

Fixes #169


mkmelin:
Screenshot looks good!

PR #175, opened 2020-03-01
Add latest version to data attribute on download pages.

jfx2006:
I'm working on adding support for Thunderbird to Pollbot (https://github.com/mozilla/PollBot)
with the goal of having something similar to https://mozilla.github.io/delivery-dashboard/.

One feature of Pollbot is it scrapes the latest Firefox version out of the various download
pages on www.mozilla.org in order to check that it matches what's provided by product-details.
The easiest way to do that is to have a data attribute somewhere in the page. To make it
easier to write the query selector, I added the id attribute on the element as well.

https://github.com/mozilla/PollBot/pull/253

### DIFF
--- a/build-site.py
+++ b/build-site.py
@@ -33,20 +33,22 @@ else:
     print 'Rendering www.thunderbird.net ' + langmsg
     # Prepare data and build main website.
     version = helper.thunderbird_desktop.latest_version('release')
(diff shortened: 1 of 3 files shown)


Sancus:
This looks fine to me at a glance, will test and deploy later today.

PR #181, opened 2020-04-08
Update link to build instructions page

rnons:
Seems https://developer.thunderbird.net/the-basics/building-thunderbird is 404 now, and the new url is
 https://developer.thunderbird.net/thunderbird-development/building-thunderbird


Sancus:
Thanks, fixed this.

PR #189, opened 2020-06-04
Upgrade packages and their hashes

dannycolin:
While building the website locally, I encountered an error mentioned that the hash for watchdog didn't match. I also saw that some packages were outdated so I decided to upgrade them with `pip-compile --upgrade --generate hashes requirements.txt`.

I though maybe PR'ing my local change could save you some time :).


Sancus:
Thanks, requirements have been recently updated so this should be OK.

PR #101, opened 2019-05-09
Update Slovak holiday calendar until 2025

chiak597:
Update Slovak holiday calendar until 2025

https://bugzilla.mozilla.org/show_bug.cgi?id=1550446

PR #191, opened 2020-06-18
Fix Slovak calendar in calendars.json

### LINKED ISSUE #190: Slovak holiday calendar's end date is not updated in the calendar list
SlovakHolidays.ics file was updated with holidays up to year 2025 in #101, but calendar.json was not updated. Because of that, webpage with calendar list still says "2009-2020" instead of "2009-2025".

chiak597:
Fix Slovak calendar in calendars.json (change end date from 2020 to 2025). ICS file was updated with holidays up to year 2025 in #101, but calendar.json was not updated. Because of that, webpage with calendar list still says "2009-2020". 

Fixes #190

### DIFF
--- a/media/caldata/calendars.json
+++ b/media/caldata/calendars.json
@@ -399,7 +399,7 @@
   {
     "country":"Slovakia",
     "filename":"SlovakHolidays.ics",
-    "datespan":"2009-2020",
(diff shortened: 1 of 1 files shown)


Sancus:
Thanks

PR #209, opened 2020-09-13
detect ChromeOS and make it not supported (instead of thinking it's W…

### LINKED ISSUE #153: Chrome OS should always show as unsupported
x86 Chrome OS user agents show a Windows download button on the site, which is not correct: They should show up as unsupported.

Here is an example test user agent: ```Mozilla/5.0 (X11; CrOS x86_64 11895.118.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.159 Safari/537.36```

justdave:
detect ChromeOS and make it not supported (instead of thinking it's Windows) fixes #153

### DIFF
--- a/assets/js/base/site.js
+++ b/assets/js/base/site.js
@@ -11,6 +11,10 @@
             pf = (pf === '') ? '' : pf || navigator.platform;
             ua = ua || navigator.userAgent;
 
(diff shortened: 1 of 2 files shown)


Sancus:
This works, thanks!

PR #211, opened 2020-09-21
Update screenshots to 78

Alecaddd:
I created the templates in Figma as a cloned component with different window decorations and Font families to emulate the different OSs.
I'm gonna set this as a Draft PR since I've exported only the English version.

Let me know if these look good, especially if the Windows version looks legit.
Once these are approved, I'll create all the language variations with a Figma plugin.


Sancus:
It's close but the fonts don't look right to me.

image

Alecaddd:
Apparently the fonts we list in our design guidelines are incorrect.
https://style.thunderbird.net/visuals/typography.html

> Also the header text "Subject", "Correspondents" etc is definitely too small.

Ah, you're right, I'll fix that.

I see your Windows chrome is grey instead of pale blue. Is that your customization or the default OS style?

Alecaddd:
I found out the that regular version of with Windows font I was using was incorrect.
Replaced it, fixed some sizing and colours, and updated all the screenshots.

I'd like to push this, if it's correct, and then batch replace these screenshots in English for all the other languages, in order to offer an updated visual screenshot even if not translated.

Then, I'll take care of exporting the different languages in chunks, what do you think @Sancus ?

PR #222, opened 2020-11-21
Update for Python 3 compatibility (still Python 2.7 compatible)

justdave:


justdave:
When running in python3, the attributes on some of the html tags wind up in a different order, so a diff against production will show changes, but the same tags are all there, they just aren't in the same order.  Probably a difference in how named lists are stored.

justdave:
There's a companion PR for this in thunderbird-notes - both should probably land at the same time
https://github.com/thundernest/thunderbird-notes/pull/11

jfx2006:
As of Python 3.6, dictionaries are preserving insertion order as an implementation detail. In Python 3.7 it was made an official part of the language spec. In Python2 it could vary apparently.


> When running in python3, the attributes on some of the html tags wind up in a different order, so a diff against production will show changes, but the same tags are all there, they just aren't in the same order. Probably a difference in how named lists are stored.

Sancus (approved):
Looks good. I tested this and it seems fine, but will merge in the morning just in case.

PR #231, opened 2021-01-04
Add release channel to pre-footer download block. Fixes #158

### LINKED ISSUE #158: Beta download on every page, but release button isn't
So, because the beta button is in the "pre-footer", it ends up being rendered on many pages, including the Features page, any release notes, etc.

This is a bit odd, especially on the release notes pages, because it means these pages have a beta download button, but no release channel download button. It feels like we're making it harder to find the release download than the beta download.

People have [reported the lack of release download as a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1605524). I think we could go in two directions with this: Somehow add the release download button, or change the way the footer thing works. Or maybe it's fine as-is? It's definitely weird, though.

@Alecaddd @ryanleesipes What do you guys think about this?

justdave:
Shows the beta tab by default on the index page, and the release tab by default everywhere else.

Put "Channel:" in front because there wasn't enough space for all three of them to say "channel" afterwards.


justdave:
Suggestions welcome for the text in the release tab... I swiped the string that was above the download box on the front page to avoid having to add a new string to be localized. If that's good enough, then it's good.

justdave:
The release.png and release-high-res.png images are copies of the release-100.png and release-200.png files that were already in the same directory, but the currently-used image macro needed those filenames.

Sancus (changes requested):
This seems mostly OK, I really dislike the "Channel:" header though. I think we should go with "Release Channel", "Beta Channel", "Nightly Channel".

The reason is that adding an unclickable section to the header makes the buttons smaller and rather fiddly to click, especially the Beta one. And as far as I can tell, there is no width issue above 480px wide. Which is not a good reason to make a design decision: we can alter the layout with a media query.

PR #239, opened 2021-03-19
removed days that are not holidays in Poland

mik2:
The calendar contained various days that were not actual holidays in Poland (Dzień Babci/Dziadka/Matki/Ojca/Dziecka).
IMO they should not belong to this calendar, so I have removed them.


mik2:
Also have changed the form of holiday category from plural (Święta) to singular (Święto).

PR #246, opened 2021-04-27
Complete rewrite of organizations-index.html

thomasd8:
Flesh out landing page for organizations/enterprise with more reasons for using Thunderbird and add descriptions and links for more resources, esp. enterprise.thunderbird.net.


website/organizations/index.html
```diff
@@ -22,14 +22,79 @@ <h3 class="header-section">
...
+
+        <h3 class="subheader-section">
+          {{ _('Stability, efficiency, and security for your emails') }}
+        </h3>
+        <p class="font-lg tracking-wider leading-loose mb-10 mt-0 p-links-blue">
+          {{ _('Provide your team with a stable communications solution which allows them to focus on getting their work done. The <a href="/">regular Thunderbird release</a> follows the so-called <a href="{firefox_release_calendar}">Extended Support Release (ESR) schedule</a> of the Mozilla platform. So to your advantage it will remain stable and feature-stable on that major version for approximately one year. Of course, we&#39;ll always keep you on the safe side with monthly minor upgrades which may contain stability, correctness and security upgrades.')|f(firefox_release_calendar=url('firefox.release-calendar')) }}
```
Sancus, line 31: I would take out `so-called`. Makes this phrase seem dismissive and/or sarcastic.

This whole paragraph seems kind of repetitive and it's also wrong. We've done minor updates twice a month since 78, and we also have been releasing features into those minor updates...I actually think the old paragraph was better and didn't make any already-broken promises...

I'm not sure what this new paragraph is adding other than length?

website/organizations/index.html
```diff
@@ -22,14 +22,79 @@ <h3 class="header-section">
...
+            {{ _('Integrated calendar application (PIM)') }}
+          </li>
+          <li>
+            {{ _('Integrated chat client') }}
+          </li>
+          <li>
+            {{ _('Advanced Search and Filter tools') }}
+          </li>
+          <li>
+            {{ _('Inbuilt message encryption (OpenPGP, S/MIME)') }}
```
Sancus, line 74: Strictly speaking, this should be `Built-in` for "en-US" english.

website/organizations/index.html
```diff
@@ -22,14 +22,79 @@ <h3 class="header-section">
...
+
+        <ul class="font-md tracking-wider leading-loose mb-6 p-links-blue">
+          <li>
+            {{ _('For Windows deployment, you can find the Thunderbird MSI Installer on the <a href="{syslang}">Systems &amp; Languages page</a>.')|f(syslang=url('thunderbird.latest.all')) }}
+          </li>
+          <li>
+            {{ _('For more information about deploying and configuring Thunderbird in business and enterprise environments, please visit our enterprise deployment portal: <a href="{thunderbird_etn}">enterprise.thunderbird.net</a>.')|f(thunderbird_etn=url('thunderbird.enterprise.documentation')) }}
+          </li>
+          <li>
+            {{ _('You can also join the <a href="{tb_list_enterprise}">Thunderbird Enterprise mailing list on Topicbox</a> to connect with other administrators and share advice on Thunderbird deployment (Please note: The <a href="tb_oldlist_enterprise">old tb-enterprise mailing list</a> has been retired).')|f(tb_list_enterprise=url('thunderbird.lists.enterprise'), tb_oldlist_enterprise=url('thunderbird.oldlists.enterprise')) }}
```
Sancus, line 49: Do we really want to be linking to the old mailing lists? Is there important stuff people might need to search for in there? I feel like we shouldn't be mentioning outdated resources.

website/organizations/index.html
```diff
@@ -22,14 +22,79 @@ <h3 class="header-section">
...
+            {{ _('Security, privacy, data ownership') }}
+          </li>
+          <li>
+            {{ _('Designed for efficiency') }}
+          </li>
+          <li>
+            {{ _('Implements accessibility standards') }}
+          </li>
+          <li>
+            {{ _('Open source code - transparency without proprietary mysteries') }}
```
Sancus, line 92: I like this phrase.

Sancus (changes requested):
Looks good other than a few nitpicks and minor technical issues! Thanks for doing this.

website/organizations/index.html
thomasd8: Magnus and I both didn't like that the old paragraph was diverting attention to Firefox and per old wording started out with what we do *not* have.
To address your concerns, I've simplified this paragraph, removed the broken promises, and added back the sentence about beta.

website/organizations/index.html
```diff
@@ -22,14 +22,79 @@ <h3 class="header-section">
           <span>{{ svg('envelope') }}</span> {{ _('Thunderbird for Organizations')}}
         </h3>
         <p class="font-lg tracking-wider leading-loose mb-6">
-          {{ _('Unlike the <a href="{firefox_esr}">Mozilla Firefox ESR</a>, Thunderbird does not provide a separate ESR release. This is because the <a href="/">regular Thunderbird release</a> is already an ESR release. It remains stable and on that major version for approximately one year. Only <a href="/#channel">Thunderbird Beta</a> follows the more frequent major version releases.')|f(firefox_esr=url('firefox.enterprise')) }}
+          {{ _('Thunderbird comes with a number of key benefits which make it an excellent choice for organizational deployments of any size. And it&#39;s more than just email: As an all-in-one communication platform with integrated calendar and chat, Thunderbird is all about efficiency. Whether you are a small business or organization, a government institution, or a large enterprise - Thunderbird has been designed to scale to your needs. Learn more about the <a href="#benefits">benefits of using Thunderbird in an organization.</a>')}}
```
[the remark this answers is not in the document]
thomasd8: thanks, good catch

website/organizations/index.html
thomasd8: Afasik there IS important stuff in there, but maybe you're right, this isn't the right place to refer to collections of  wisdom from the past.

website/organizations/index.html
thomasd8: Perfect!

website/organizations/index.html
thomasd8: Thanks 😃

thomasd8:
Thanks for the review! I've addressed all of the review comments and added new commits; let me know if that's the right thing to do wrt github workflows.

Sancus:
> 
> 
> Thanks for the review! I've addressed all of the review comments and added new commits; let me know if that's the right thing to do wrt github workflows.

Yeah, adding commits is good! Makes it easy to see changes and they all get squashed down to 1 by the final merge anyhow.

website/organizations/index.html
Sancus: Fair enough. I apologize if that was a bit strongly worded. I'm a fan of brevity where possible. I think the new paragraph works well!

Sancus:
Thanks for this update!

PR #250, opened 2021-05-10
fix markupsafe would not install, builder watch error with python 3, and add IndiaHolidays to calendars.json

Zahrun:


Sancus:
This seems to be missing the actual updated calendar file.

Sancus (changes requested):
Needs the actual calendar file.

Sancus:
Thanks, but no longer needed.

PR #260, opened 2021-06-29
Update Singapore Holidays

### LINKED ISSUE #252: Singapore Public holidays for 2021 and 2022
These are the ICS files for Singapore public holidays in 2021 and 2022
It would be good to have them accessible in TB
[public-holidays-sg-202x.zip](https://github.com/thundernest/thunderbird-website/files/6471543/public-holidays-sg-202x.zip)

justdave:
fixes #252


justdave:
See comments in #252 before reviewing...

Sancus:
I don't think we should pull the files in like this. When the issue was initially filed, it wasn't clear that it was a dynamic government source. Since that source exists, I think we should link directly to it(current year only) dynamically so nobody needs to file a bug for this country again.

Sancus (changes requested):
See previous comment.

mkmelin:
Linking to 3rd party directly has (some) privacy implications.

Sancus:
> Linking to 3rd party directly has (some) privacy implications.

Sure, but I don't think that's a big deal for a calendar download. I guess we could write a script to automatically pull the calendar file from any URL we specify at build time if it's not already there for the current year, not sure it's worth it though. Perhaps there are other calendars we could auto-update from external sources as well.

I do know for a fact that people will need to re-file this bug every year for us to update this calendar otherwise, and any calendar we can automatically update we should because these bugs waste time.

mkmelin:
Well it's not a download. You subscribe to it and it will be polled every <poll interval>

Sancus:
> 
> 
> Well it's not a download. You subscribe to it and it will be polled every

They have separate ics files for each year eg https://www.mom.gov.sg/-/media/mom/documents/employment-practices/public-holidays/public-holidays-sg-2021.ics and then -2022.ics etc.

So I assume they are NOT treating it as a dynamic calendar that will be updated.

justdave:
@mkmelin / @Sancus So what's the verdict here?

Sancus:
It's not worth spending more time on this, I'm taking it as-is.

PR #261, opened 2021-06-29
change Nightly to Daily where possible.

### LINKED ISSUE #247: Update the wording from Nightly to Daily
If I'm not wrong, `Nightly` is used for Firefox while `Daily` is used for Thunderbird.
I think we should update the wording around the website to reflect this aspect, especially in the pre-footer tabs to download the different versions.

We should also update the icons to use one with the "daily" text on it, like we currently use in the builds: https://searchfox.org/comm-central/rev/faba8b4037ce7434416ee2f5fb2d00497675795b/mail/branding/nightly/content/about-logo.svg

justdave:
Fixes #247


Sancus:
The code here is good, but lets not merge it until we figure out what's going on with the icon coloring.

justdave:
Branch updated to include the new logos that are all the same color.

Sancus (approved):
This works, l10n updates are required though so I'll merge it tomorrow.

PR #262, opened 2021-06-29
make calendar rows always resize to fit.

### LINKED ISSUE #230: Calendar Index page requires horizontal scrolling if the browser window is between 1024 and 1216 pixels wide
https://www.thunderbird.net/en-US/calendar/holidays/

If the browser window is less than or equal to 1024 pixels wide, the width of the chart is set to match the width of the window (good).

If the browser window is greater than 1024 pixels wide, the width of the chart is set to 1216 pixels, which requires horizontal scrolling until the window size exceeds that.

The dynamic width should remain in place up to the max width rather than jumping to the max when you hit 1024 width.

justdave:
fixes #230

### DIFF
--- a/website/calendar/holidays/index.html
+++ b/website/calendar/holidays/index.html
@@ -44,7 +44,7 @@
     </ul>
   </section>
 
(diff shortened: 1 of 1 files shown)


Sancus (approved):
Looks good, thanks!

Issue #238, opened 2021-03-09
wordmark svg in footer looks broken at viewport widths from 768 to 1023px


Sancus:
![image](https://user-images.githubusercontent.com/479671/110550317-48f95580-80e8-11eb-8740-43901a3b8257.png)

The svg in question is https://github.com/thundernest/thunderbird-website/blob/master/media/svg/logo-wordmark-white.svg

Not too sure how CSS can break rendering of an svg like this tbh, I poked at it a bit and I have no idea what is going on.

PR #267, opened 2021-08-09
Fix broken responsive behaviour of logos (#238)

sebastianlay:
The issue is that the `logo-wordmark-white.svg` is embedded twice on the page (once in the header, once in the footer). The gradients that are defined in the SVG (`id="linearGradient-1"` and so on...) are then also defined twice. For this breakpoint the SVG in the footer referenced the gradients that were defined in the (now hidden) SVG in the header. Not hiding this logo fixes the issue:

footer

Additionally this also fixed the broken header logo for this breakpoint, which is caused by the SVG having an inline width and height, which doesn't get reset by CSS (via `svg { width: 100%; height: auto; }` for example):

header

### DIFF
--- a/website/_includes/base-resp.html
+++ b/website/_includes/base-resp.html
@@ -17,9 +17,8 @@
     <section class="flex justify-between items-center p-6 flex-wrap">
       {% block site_header_logo %}
         <h2 class="mt-0 mb-0">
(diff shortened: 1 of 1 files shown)


Alecaddd (approved):
Good finding, and thanks for the fix.

Sancus:
Thanks very much for this fix!

PR #275, opened 2021-10-13
Fix RTL Sidebar

JustE3saR:
I have hidden the horizontal sidebar that appears if the website is right to left.

### DIFF
--- a/media/css/normalize.css
+++ b/media/css/normalize.css
@@ -17,11 +17,13 @@ html {
    ========================================================================== */
 
 /**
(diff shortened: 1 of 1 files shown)


Sancus:
Hi, I appreciate the bug report, I don't think this is the correct fix though. For one thing, that css file is an external library.

Instead I've modified the offending bit of anti-bot html directly: https://github.com/thundernest/thunderbird-website/commit/d912888023b948b1deb7db2eda977de23c6b7f06

PR #280, opened 2021-12-04
Updating Hungarian Holidays

tozo:
Updating Hungarian Holidays
Making fixed holidays to repeat annually
The non fixed holidays are updated up until 2060


Sancus:
Thanks.

PR #292, opened 2022-03-15
Use matrix.to links to link to matrix channels

freaktechnik:
When Mozilla first started using Matrix there was no good way for matrix.to to guide people to chat.mozilla.org, which is why usually chat.mozilla.org deep links were used. However, matrix.to now allows specifying the element web client instance that should be advertised. Using matrix.to links has the benefit that it can offer direct links for opening the room in other clients, or instructions, or on Android they are usually the main link handler Matrix apps will register. Thus using matrix.to links makes it easier for users that already are in the Matrix ecosystem to join the rooms, while still offering a simple path to the element web instance.

### DIFF
--- a/website/contact/index.html
+++ b/website/contact/index.html
@@ -44,10 +44,10 @@
 
       <article class="flex flex-col p-4 flex-1 lg:max-w-sm mb-10 lg:mb-0 bg-white rounded shadow-lg">
         <h4 class="mt-0 mb-4 uppercase text-blue font-md">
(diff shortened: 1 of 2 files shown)


clokep:
Links seem to be reasonable to me! 👍

Sancus:
I wasn't totally sure I wanted to approve this since it does mean sending people through a non-Mozilla server, but meh, it's only for Matrix so not really that big of a deal.

PR #296, opened 2022-04-13
Donation Modal Style

devmount:
This change revamps the markup and adds styles to the donation modal according to the given mockup.

image

- A close button (x) in the upper right corner was added
- The donation amount selection was implemented - the box with the corresponding radio button gets highlighted when clicked
- Clicking the "Other amount" input field automatically activates the radio button. Also the value of the radio button gets updated on input.
- Responsive and IE11 compatible

Not in the scope of this PR but planned for the next changes:

- Updating currency symbol in presets when currency is changed
- Making CTA buttons work
- URLs for the links in the modal footer

### DIFF
--- a/assets/less/components/donation-elements.less
+++ b/assets/less/components/donation-elements.less
@@ -1,18 +1,49 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 .modal {
-    width:50%;
-    margin-left:auto;
-    margin-right:auto;
-    padding: 5px;
-    border: thin #000 solid;
-    background-color:#fff;
-    z-index:3; /* places the modal dialog on top of everything else */
-    position:fixed;
-    top:25%;
-    left:25%;
-    display:none;
+    width: 100%;
+    max-width: 640px;
+    border-radius: 5px;
+    box-shadow: 0 5px 15px -10px rgb(0 0 0 / 60%), 0 0 3px rgb(0 0 0 / 10%);
+    background-color: #fff;
+    z-index: 3; /* places the modal dialog on top of everything else */
+    position: fixed;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    display: none;
+
+    #close-modal {
+        position: absolute;
+        z-index: 1;
+        top: 0;
+        right: 0;
+        padding: .25rem;
+        margin: .5rem;
+        cursor: pointer;
+
+        svg {
+            height: 20px;
+            width: 20px;
+            stroke-width: 2.5;
+            stroke: #bebebe;
+            fill: none;
+            stroke-linecap: round;
+            stroke-linejoin: round;
+        }
+        &:hover svg {
+            stroke: #000;
+        }
+    }
+
+    h1 {
+        color: #245cad;
+        text-align: center;
+        font-size: 3em;
+        font-weight: 300;
+    }
 }
 
 #modal-overlay {
--- a/assets/less/components/donation-elements.less
+++ b/assets/less/components/donation-elements.less
@@ -4,25 +4,17 @@
 
 .modal {
     width: 100%;
-    max-width: 640px;
-    border-radius: 5px;
-    box-shadow: 0 5px 15px -10px rgb(0 0 0 / 60%), 0 0 3px rgb(0 0 0 / 10%);
-    background-color: #fff;
-    color: #333;
     z-index: 3; /* places the modal dialog on top of everything else */
-    position: fixed;
-    top: 50%;
-    left: 50%;
-    transform: translate(-50%, -50%);
+    top: calc(50% - 250px);
+    margin: 0 auto;
+    left: 0;
+    right: 0;
     display: none;
 
     #close-modal {
-        position: absolute;
         z-index: 1;
         top: 0;
         right: 0;
-        padding: .25rem;
-        margin: .5rem;
         cursor: pointer;
 
         svg {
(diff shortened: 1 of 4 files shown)


website/_includes/donate-modal.html
```diff
@@ -2,39 +2,116 @@
...
+      <p>{{ _('Thunderbird is independent, free, and Open Source software sustained by your donations. Help us keep the project alive!') }}</p>
+    </section>
+    <section class="amount">
+      <div class="flex justify-between items-center">
+        <label class="uppercase font-sm">{{ _('Select an amount') }}</label>
+        <label>
+          {{ _('Currency') }}
+          <select name="currency-switcher-currency" id="id_currency-switcher-currency">
+            {% for currency, vals in settings.CURRENCIES | dictsort %}
+            <option value="{{ currency }}" {% if currency == this_currency %}selected{% endif %}>{{ currency.upper() }}   {{ vals.symbol }}</option>
```
Sancus, line 34: None of this actually runs due to the removal of the `.decode('utf-8')` part -- it was there for a reason :)

This might be a Python2 vs 3 issue if it was working for you, the server side site build is still using Python2. I've been meaning to correct this, can probably do it on Monday.

devmount: Oh I'm sorry, I was indeed using Python 3.6 for development and forgot to put those decode statements back.

Both would be fine for me (developing in Python 2 or waiting for you to upgrade to Python 3) - let me know what works best for you. I won't work on this before Monday anyway.

Sancus: Yeah lets deal with this on Monday.

Alecaddd:
Thank you so much for working on this.
I'll review this on Monday as well, sorry for the delay.
Pulling in @micahilbery as our newly hired designer, so he can start getting exposed to the website work.

website/_includes/donate-modal.html
Alecaddd: Is this related to this error?
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 1: ordinal not in range(128)`
I can't build it locally.

Alecaddd (changes requested):
Overall it looks nice and it's visually going towards the right direction.

Unfortunately the CSS is not correct the way it has been written.
I know this will feel a bit weird, and I apologize, but we have a modular, utility-first, auto-generating CSS that we use in order to guarantee visual consistency and components reusability.

All the hard coded sizing, colors, spacing, etc, are not a good approach since we already have all those defined in reusable classes that should be added via the `&:extend()`, or directly added as classes to the HTML element.

This will help also to make this new element properly responsive by leveraging existing breaking points and globally consistent sizes.

I know the utility-first approach is a controversial topic, but it's fundamental for us to stick with it due to external contributors modifying the site. If we don't do that, we risk to end up with repeated classes and hardcoded values, making it very hard to keep the visuals consistent.

I'm more than happy to help you out by suggesting code edits and classes to make sure the donation dialog uses our built-in styling.

devmount:
@Alecaddd Many thanks for the feedback! 👏🏻 

> I know the utility-first approach is a controversial topic, but it's fundamental for us to stick with it due to external contributors modifying the site.

No need to apologize, I totally understand that and will try to adopt it.

> All the hard coded sizing, colors, spacing, etc, are not a good approach since we already have all those defined in reusable classes that should be added via the `&:extend()`, or directly added as classes to the HTML element.

Am I allowed to add sizes, colors etc.? Or should I only use existing definitions?

website/_includes/donate-modal.html
devmount: @Alecaddd Yes it is. I made it Python2 compatible now - you should be able to build it again.

devmount:
@Alecaddd I made some changes and tried to use the built-in styling as much as possible. Please review again and see, if it's correct now. Note that I added two colors and uncommented some existing utility classes in `variables.less`.

Alecaddd:
> @Alecaddd I made some changes and tried to use the built-in styling as much as possible. Please review again and see, if it's correct now. Note that I added two colors and uncommented some existing utility classes in `variables.less`.

Thank you so much, I'll review it in a bit.

> Am I allowed to add sizes, colors etc.? Or should I only use existing definitions?

Oh yes, absolutely. Feel free to add variables and all the definitions you need.

website/_includes/donate-modal.html
```diff
@@ -2,39 +2,116 @@
...
+  {% set this_currency = settings.LOCALE_CURRENCIES[LANG] %}
+{% else %}
+  {% set this_currency = 'usd' %}
+{% endif %}
+{% set this_curdata = settings.CURRENCIES[this_currency] %}
 
-<div id="amount-modal" class="modal">
-<h1>{{ _('Pay What You Want') }}</h1>
-<p>{{ _('Thunderbird is independent, free, and open source software sustained by your donations. Help us keep the project alive!') }}</p>
+<div id="amount-modal" class="modal rounded w-full max-w-2xl shadow-lg bg-white text-black-lighter fixed z-30 mt-0 mx-auto pin-x hidden">
```
Alecaddd, line 12: nit: I would use a `rounded-lg` here as I think it looks nicer.

assets/less/components/donation-elements.less
```diff
@@ -1,30 +1,49 @@
...
+                &:hover {
+                    &:extend(.border-blue);
+                }
+                &:last-child {
+                    &:extend(.pt-0, .pb-0, .pr-0, .bg-grey-lighter);
+                    flex-basis: @widths[w-40];
+                    #amount-other {
+                        &:extend(.w-full, .h-full, .pl-2, .border-0, .border-l, .border-solid, .border-grey, .font-xl);
+                        box-sizing: border-box;
+                    }
```
Alecaddd, line 34: Let's style the placeholder with a smaller font size so it looks nicer.

```suggestion
                    }
                    
                    &::-moz-placeholder,
                    &::-webkit-input-placeholder,
                    &::placeholder {
                        &:extend(.font-md);
                    }
```

website/_includes/donate-modal.html
```diff
@@ -2,39 +2,116 @@
...
-    <button id="amount-cancel">{{ _('No Thanks; Close & Download') }}</button>
+    <section class="font-lg pt-0 pl-6 pb-4 pr-6 leading-normal text-center">
+      <p>{{ _('Thunderbird is independent, free, and Open Source software sustained by your donations. Help us keep the project alive!') }}</p>
+    </section>
+    <section class="bg-grey-light pt-4 pl-6 pb-4 pr-6">
+      <div class="flex justify-between items-center">
+        <label class="uppercase font-sm">{{ _('Select an amount') }}</label>
+        <label>
+          {{ _('Currency') }}
+          <select id="id_currency-switcher" class="inline-block w-24 bg-white p-1 border border-solid border-grey" name="currency-switcher">
```
Alecaddd, line 32: We can add a `rounded` class so this looks nicer.

website/_includes/donate-modal.html
```diff
@@ -2,39 +2,116 @@
...
+            {% for currency, vals in settings.CURRENCIES | dictsort %}
+            <option value="{{ currency }}" {% if currency == this_currency %}selected{% endif %}>{{ currency.upper() }}   {{ vals.symbol.decode('utf-8') }}</option>
+            {% endfor %}
+          </select>
+        </label>
+      </div>
+      <div id="amount-selection" class="amount-selection flex flex-wrap mt-3">
+        <label class="flex items-center shadow-md mb-2 mr-4">
```
Alecaddd, line 40: We can add a `rounded` class to all this value selectors. The last one with the amount needs an `overflow: hidden` so the input field inside doesn't bleed out.

assets/less/components/donation-elements.less
```diff
@@ -1,30 +1,49 @@
...
-    padding: 5px;
-    border: thin #000 solid;
-    background-color:#fff;
-    z-index:3; /* places the modal dialog on top of everything else */
-    position:fixed;
-    top:25%;
-    left:25%;
-    display:none;
-}
+    top: calc(50% - 250px);
```
Alecaddd, line 6: I would use
```
top: 50%;
transform: translateY(-50%);
```
So it's vertically centered in the page and if the browser height is too narrow, the important content is still visible at the center.

Alecaddd (approved):
This is great, awesome job!
I left some small nit comments for some purely visual changes, nothing major.

A couple of notes for future PRs:
- Pressing Esc should close the dialog
- Clicking outside the dialog (on the overlay) should close it
- When the dialog opens, an `overflow: hidden` style should be added to the body so the content doesn't scroll underneath the dialog

devmount:
@Alecaddd Many thanks for your valuable feedback! I've implemented all your suggestions.

Alecaddd:
@Sancus This looks good, is it okay to squash and merge?
It goes into the `new` branch so I guess it won't be published on the website.

PR #299, opened 2022-04-18
Display bug links for beta releases only.

jfx2006:
Wayne asked to display bug links for beta releases only. This is a trial to see how it works out.


wsmwk:
Looks good. Thanks

PR #306, opened 2022-05-10
Add anti-discrimination statement

KillYourFM:
It's present beneath the last two open positions, but not the first one. This creates confusion and we've had some people express concern over that.


rnayabed:
समाज

Sancus:
This is just a mistake. I think the original intent was to remove the repetition in every job description and put it in the How to Apply section, but this is fine too.

Certainly there was never any intent that the anti-discrimination statement could apply to specific jobs but not others.

PR #307, opened 2022-05-10
Remove references to Lightning on main page

thomasd8:


Sancus:
Thanks @thomasd8 !

This is good; just didn't want to merge it until I was done pushing out the new beta page. This will break the l10n for those strings so it needs to sit for a bit before it goes out.

PR #311, opened 2022-05-23
Add Beta CTA to banner

### LINKED ISSUE #302: feature beta download choice more prominently on thunderbird.net
We will want issue #301 - update beta start page - to be live before the download change is made live.

micahilbery:
fixes #302

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -16,6 +16,12 @@
     <div class="tracking-wider">
       {{ download_thunderbird(force_direct=true, alt_copy=_('Free Download')) }}
     </div>
+    <div class="flex flex-wrap items-center justify-center pt-6">
+      <p class="font-md p-3 m-0">
+        {{ _('<span class="font-semibold">Want to try new features early?</span> Get the beta and be a part of the future of Thunderbird!') }}
+      </p>
+      <a href="https://support.mozilla.org/kb/thunderbird-beta" class="btn-beta rounded-full pt-2 pb-2 pr-3 pl-3 mb-0">{{ _('Get the Beta') }}</a>
+    </div>
   </section>
   <section class="hidden md:flex max-w-6xl self-center">
     {{ platform_img('thunderbird/landing/screenshot.png', {'alt': _('Thunderbird screenshot'), 'high-res': True, 'l10n': True, 'platforms': ('windows', 'linux', 'mac')}) }}
--- a/website/index.html
+++ b/website/index.html
@@ -20,7 +20,7 @@
       <p class="font-md p-3 m-0">
         {{ _('<span class="font-semibold">Want to try new features early?</span> Get the beta and be a part of the future of Thunderbird!') }}
       </p>
-      <a href="https://support.mozilla.org/kb/thunderbird-beta" class="btn-beta rounded-full pt-2 pb-2 pr-3 pl-3 mb-0">{{ _('Get the Beta') }}</a>
+      <a href="https://support.mozilla.org/kb/thunderbird-beta" class="btn-beta btn-beta-pill">{{ _('Get the Beta') }}</a>
     </div>
   </section>
   <section class="hidden md:flex max-w-6xl self-center">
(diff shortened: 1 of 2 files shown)


website/index.html
```diff
@@ -16,6 +16,12 @@ <h1 class="font-4xl md:font-hero font-thin md:mt-10 mb-8">{{ self.page_title() }
     <div class="tracking-wider">
       {{ download_thunderbird(force_direct=true, alt_copy=_('Free Download')) }}
     </div>
+    <div class="flex flex-wrap items-center justify-center pt-6">
+      <p class="font-md p-3 m-0">
+        {{ _('<span class="font-semibold">Want to try new features early?</span> Get the beta and be a part of the future of Thunderbird!') }}
+      </p>
+      <a href="https://support.mozilla.org/kb/thunderbird-beta" class="btn-beta rounded-full pt-2 pb-2 pr-3 pl-3 mb-0">{{ _('Get the Beta') }}</a>
```
Alecaddd, line 23: I don't think we want to link the support article here.
@wsmwk Should this link scroll the page to the bottom or directly trigger the beta download + donation page?

wsmwk: @Alecaddd @ryanleesipes and I agreed to send this to the support page. This is my strong opinion that users be presented with information prior to installing beta, because users are putting important data at risk, not equivalent to Firefox beta.

Maybe at some point in the future when beta is more reliably stable, etc, we could reconsider this. 

Given that we took a different approach to implementing issue #302  the reference at the bottom should also be changed to point to the support page.

Triggering donations flow is going beyond the scope of this project IMO.

wsmwk (approved):
I think this is a solid choice. LGTM

website/index.html
Alecaddd: Sounds good, thanks for the explanation.

Alecaddd (changes requested):
We need to tweak a bit the style to make it play more nicely with the adjacent content.
Right now this pill button is very large and immediately steals the attention. The gradient and shape of the button are already enough eye catching and pretty impossible to miss, so we need to reduce a bit its sizing to maintain a good visual hierarchy.

Let's add this class variation to the CSS `.btn-beta`:
```
&.btn-beta-pill {
    &:extend(.rounded-full, .pt-1, .pb-1, .pr-2, .pl-2, .mb-0, .tracking-normal, .font-regular, .font-bold, .leading-normal);
  }
```
So you can simply add the `btn-beta-pill` class to the HTML link instead of writing all those attributes inline.

Before:
Screenshot from 2022-05-23 19-07-13

After:
Screenshot from 2022-05-23 19-07-20

Alecaddd (approved):
Looks good, thanks!

Alecaddd:
@Sancus gentle ping for Andrei to make him aware of this change landing on `master`.
Let me know when this can be deployed to production, after maybe a quick check on stage.

Sancus:
I really don't like the way this has turned out:
- We've had people download beta for years via one-click on the page. Now it needs a whole support article of warnings and a convoluted, broken download flow? Why?
- OK, you want to add information. That's fine. Why is it in a support article that has a "download button" which is just a link and which ruins the whole experience of the website?

The page with information about beta should really be on the website so that the download button can then work normally. Asking people to read a page of text and THEN go and hunt for their build on the beta-all page makes downloading beta actually much harder than it was in the past.

Is the goal to drive more people to beta or not? Because this doesn't seem like a good way to do that.

Sancus:
Alright, I've added a download landing page for beta to the website. It's not perfect, but at least it prevents the unfriendly download flow.

https://www-stage.thunderbird.net/en-US/download/beta/

Feel free to comment. I did make some copy edits as well, and I removed the Get Involved call at the end. We already have links to that elsewhere on the website, and this page is too long as it is, so anything we can cut is an improvement.

ryanleesipes:
Great job @Sancus - we also need to plan to sunset the beta CTA when 102 is released. We should probably already have a pull request ready to do that @micahilbery

PR #321, opened 2022-06-27
Update windows screenshots

micahilbery:
After some discussion, I have changed the screenshots to better reflect the default spaces toolbar and density.


micahilbery:
Screenshot 2022-06-27 at 18-31-13 New in Thunderbird 102

Alecaddd:
Only one thing for me to change.
In the header image, the light version should have its won drop shadow to detach it from the white message body of the dark mode.
At first glance it seems like a visual glitch.

Alecaddd:
@micahilbery maybe the dark screenshot in the header could show a mult threaded message selected, so it's fully dark and we don't need any drop shadow.

micahilbery:
Screenshot 2022-06-27 at 19-47-18 New in Thunderbird 102

micahilbery:
I went the drop shadow route because there isn't a message thread in the nemo account.

Alecaddd (approved):
Looks good, thanks!

PR #324, opened 2022-07-05
Peru Holidays

pitfiend:
This file has the official Peruvian holidays, except for Easter Sunday. The holidays are yearly repite and fixed on their according day.

[PeruHolidays.zip](https://github.com/thundernest/thunderbird-website/files/9049612/PeruHolidays.zip)


Sancus:
We're not going to do calendars this way anymore, see #340.

PR #328, opened 2022-08-17
Set known esr major release versions eplicitly to avoid mishaps.

### LINKED ISSUE #325: Links missing/broken for some 102.x releases on release notes overview page
I noticed that the links for the 102.1.0 and 102.1.1 releases are missing on https://www.thunderbird.net/en-US/thunderbird/releases/ . Then I scrolled down and found them listed in the 10.0 release series, the links for the 102.0.1/102.0.2/102.0.3 releases are also listed there. (As a result, these links for the 102 series are broken.)

jfx2006:
Fixes #325.

The 102.x release notes were being thrown off by "esr_major_versions" not including
102. Rather than try to define a pattern for the esr major versions, adopt the method
used by the "mozilla-version" Python package, and just explicitly liat them out.
This does mean adding to the list every year, but there's no chance of getting
thrown off when the interval changes again.

### DIFF
--- a/product_details.py
+++ b/product_details.py
@@ -143,12 +143,7 @@ class ThunderbirdDetails():
         return self.platform_labels.items()
 
     def list_releases(self, channel='beta'):
-        version_name = self.version_map.get(channel, 'LATEST_THUNDERBIRD_DEVEL_VERSION')
-        esr_major_versions = (
-            list(range(10, 59, 7))
-            + [60, 68]
-            + list(range(78, int(self.current_versions[version_name].split('.')[0]) + 1, 13))
-        )
+        esr_major_versions = (10, 17, 24, 31, 38, 45, 52, 60, 68, 78, 91, 102)
         releases = {}
         for release in self.major_releases:
             major_version = float(re.findall(r'^\d+\.\d+', release)[0])


jfx2006:
For reference, mozilla-version: https://github.com/mozilla-releng/mozilla-version/commit/4312f51385b612b8f07a73a0d73cfffdac1a7d31

thomasd8:
https://bugzilla.mozilla.org/show_bug.cgi?id=1786235

thomasd8:
I don't fully understand this, but if having an explicit version list which needs manual update once a year can replace a broken algorithm (which also needs manual intervention now), that sounds like a reasonable compromise.
So can someone accept/land this pull request? @Alecaddd

Sancus:
> I don't fully understand this, but if having an explicit version list which needs manual update once a year can replace a broken algorithm (which also needs manual intervention now), that sounds like a reasonable compromise. So can someone accept/land this pull request? @Alecaddd

If you don't fully understand a thing, why are you trying to tell people what to do about it? Don't do this. Just because someone filed a bug, doesn't mean something has to be fixed immediately.

The reason I haven't pushed a fix for this issue has nothing to do with this patch, I was waiting for l10n on the whatsnew page updates to do a production push. I should be able to do a push today, which is what I had planned to do before.

PR #332, opened 2022-09-20
Use official spelling of "Daylight Saving Time"

smmr0:
The official spelling is "Daylight Saving Time" (not "Daylight Savings Time"). This is the spelling that the National Institute of Standards and Technology (NIST) uses; they also [link to](https://www.nist.gov/pml/time-and-frequency-division/links-time-and-frequency-exhibits) [a website that explains in further detail](http://www.webexhibits.org/daylightsaving/b2.html):

> ## Spelling and grammar
> 
> The official spelling is Daylight Saving Time, **not** Daylight Saving**S** Time.
> 
> Saving is used here as a verbal adjective (a participle). It modifies time and tells us more about its nature; namely, that it is characterized by the activity of saving daylight. It is a saving daylight kind of time. Because of this, it would be more accurate to refer to DST as daylight-saving time. Similar examples would be a mind-expanding book or a man-eating tiger. Saving is used in the same way as saving a ball game, rather than as a savings account.
> 
> Nevertheless, many people feel the word _savings_ (with an 's') flows more mellifluously off the tongue. _Daylight Savings Time_ is also in common usage, and can be found in dictionaries.
> 
> Adding to the confusion is that the phrase _Daylight Saving Time_ is inaccurate, since no daylight is actually saved. _Daylight Shifting Time_ would be better, and _Daylight Time Shifting_ more accurate, but neither is politically desirable.

### DIFF
--- a/media/caldata/USHolidays.ics
+++ b/media/caldata/USHolidays.ics
@@ -11,7 +11,7 @@ COMMENT:	ML King Day 	            begins on 	3rd Monday of Jan
 COMMENT:	Groundhog Day	            begins on 	Feb 2
 COMMENT:	Valentine's Day	            begins on 	Feb 14	
 COMMENT:	President's Day             begins on 	3rd Monday of Feb
-COMMENT:   **	Daylight Savings            begins on 	2nd Sunday of March
+COMMENT:   **	Daylight Saving Time begins begins on 	2nd Sunday of March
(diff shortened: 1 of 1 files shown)


Sancus:
I'm going to merge this for the benefit of people on that old calendar, but calendars will now be autogenerated: https://www-stage.thunderbird.net/media/caldata/autogen/United-States-Holidays.ics

PR #335, opened 2022-11-15
EOY Appeal

micahilbery:
First pass at the EOY Appeal. I'm really struggling with our build system today. So it's fully not where I want it to be but this is a first pass. I think the typography could be improved as well as some animations and things.


Alecaddd:
It's a good start, but of course it's missing a lot of things.
If it can help, I was able to solve all the problems by manually installing python 2.7 on macOS.
https://www.python.org/ftp/python/2.7.18/python-2.7.18-macosx10.9.pkg

With that, I'm able to use the regular `pip` to install the requirements and run the watch command without problems.
Shutdown and restart are fast.

The only thing I had to change was removing [this chunk](https://github.com/thundernest/thunderbird-website/blob/master/website/_includes/bedrock/base-resp.html#L83) from the bedrock base file, otherwise it wouldn't compile for me.

ryanleesipes:
@micahilbery @Alecaddd have you guys synced up on what is missing? I'd like to get this merged today. We still have translations to do.

micahilbery:
I have it working on static local files because I can't get the build system to work. I have copied that work into the live files so it should work in theory but I cant really test it.

website/thunderbird/102.0/eoy/index.html
```diff
@@ -0,0 +1,109 @@
...
+</header>
+{% endblock %}
+
+{% block email_form %}
+{% endblock email_form %}
+
+{% block content %}
+<section class="font-xl">
+  <div class="letter-container">
+    <a href="#" class="donate-banner">
```
Alecaddd, line 56: missing donation URL.
@ryanleesipes which donation link + utm we want here?

website/thunderbird/102.0/eoy/index.html
```diff
@@ -0,0 +1,109 @@
...
+      </div>
+      <div>{{ svg('donate-heart') }}</div>
+      <div class="right-lines">
+        <div class="line"></div>
+        <div class="line"></div>
+        <div class="line"></div>
+      </div>
+    </div>
+    <p class="text-center accent-text"">
+      The Thunderbird Team
```
Alecaddd, line 102: also this

Alecaddd (changes requested):
Why the hard line on the background image? It looks nothing like the designed page.

The style of the page needs more polish and refinement, but the text is fine (after the few issues addressed), so if it's good for @ryanleesipes and @Sancus, we can merge it fo the purpose of translations, and then continue the improvements without touching the strings in a follow up PR

website/thunderbird/102.0/eoy/index.html
Alecaddd: This is the link we should add here: https://give.thunderbird.net/?utm_source=thunderbird-client&utm_medium=referral&utm_content=EOY22

website/thunderbird/102.0/eoy/index.html
```diff
@@ -0,0 +1,109 @@
...
+
+</div>
+<header>
+  {% block header_content %}
+  <section id="initAppeal" class="text-left font-xl pl-8 pr-8 md:pl-16 md:pr-16 max-w-8xl mx-auto pt-12">
+    <aside>
+      <h1 class="accent-text">
+        <span class="font-light">{{ _('Help Keep')}}</span>
+        <br>
+        <span>{{ _('Thunderbird Alive!')}}</span>
```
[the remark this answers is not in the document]
Alecaddd: I spoke with Andrei and indeed this would not be understandable or usable for a lot of languages.

Also I would suggest to style the whole H1 with the larger font we need for the "Thunderbird Alive" part, and then a simple span without class to style the exception, which is the smaller and thinner font.

We need to have something like this, I think:
```
{{ _('<span>Help Keep</span>Thunderbird Alive!')}}
```

We can also avoid the `<br>` by making that span a display block.

media/css/eoy.css
```diff
@@ -0,0 +1,367 @@
...
+  max-width: 100%;
+  line-height: 1.6;
+}
+
+@media (prefers-color-scheme: dark) {
+  body {
+    background: #204571;
+    background-position: top left;
+    background-size: auto 400px;
+    background-repeat: repeat-x;
```
Alecaddd, line 102: no need to repeat the same values for dark mode, these should be maintained from the previous body style.

Alecaddd (changes requested):

media/css/eoy.css
```diff
@@ -0,0 +1,364 @@
...
+
+.letter-container:before {
+  content: '';
+  position: absolute;
+  background: #2493EF;
+  background-image: linear-gradient(127deg, #00FFFF, #DB00FF);
+  top: -4px;
+  left: -4px;
+  right: -4px;
+  bottom: -4px;
```
Alecaddd, line 313:
```suggestion
  inset: -4px;
```

Alecaddd (approved):
This is good to land for the strings.
@ryanleesipes and @Sancus, waiting for your final review and merging decision.

ryanleesipes:
LGTM. Let's merge @Sancus

ryanleesipes (approved):
Haven't loaded up the site (reviewing from phone). But, from what I can tell - this looks good to go

PR #338, opened 2022-11-22
Refinement of EOY appeal styles

micahilbery:


media/css/eoy.css
```diff
@@ -94,152 +220,207 @@ body {
...
   background-repeat: no-repeat;
   border-radius: 3px;
+  box-shadow: 4px 4px 24px rgba(0,0,0,0.15);
   color: white;
-  display: grid;
-  grid-template-columns: repeat(2, 1fr);
-  gap: 1em;
+  display: flex;
+  justify-content: space-between;
   font-size: 1.25rem;
```
Alecaddd, line 292: Let's use a `line-height: 1.1` so the text it's tighter.

media/css/eoy.css
```diff
@@ -94,152 +220,207 @@ body {
...
-  padding: 1em;
-  margin: 2em;
-  margin-top: -75px;
+  padding: 12px;
```
Alecaddd, line 293: It's always recommended to add `padding` to the child elements of a container, otherwise it affects the defined size of the whole container.
You can see that if you change this padding the whole container grows beyond the defined size.
Let's remove this.

media/css/eoy.css
```diff
@@ -94,152 +220,207 @@ body {
...
+  margin: 0 auto;
+  margin-top: -114px;
   text-decoration: none;
-  box-shadow: var(--shadow);
+  text-shadow: var(--shadow);
   transition: transform 0.3s ease-in-out;
   position: relative;
   overflow: hidden;
+  width: 400px;
+  height: 100px;
```
Alecaddd, line 302: Without the padding, we can make this button the size we want and we can be sure it won't change.
```suggestion
  width: 430px;
  height: 120px;
```

media/css/eoy.css
```diff
@@ -94,152 +220,207 @@ body {
...
 
-.donate-banner:hover > #decoration > #pill-1 {
-  animation: 5s infinite alternate pill-float;
+.donate-banner:hover #donate-banner-left {
+  transform: translate(24px, 12px);
+}
+
+#donate-banner-left,
+#donate-banner-right {
+  transition: all .5s ease-in-out;
```
Alecaddd, line 333: we can add the padding here for these elements. `padding: 15px;`

media/css/eoy.css
```diff
@@ -94,152 +220,207 @@ body {
...
-.donate-banner:hover > #decoration > #pill-10 {
-  animation: 5s infinite alternate pill-float;
+#decoration > #pill-10 {
+  animation: 35s infinite pill-float linear;
   animation-delay: 0.4s;
 }
 
 #decoration {
   position: absolute;
+  left: -70px;
```
Alecaddd, line 397: I'd use -59px so the big pill is centered on the edge.

media/css/eoy.css
```diff
@@ -94,152 +220,207 @@ body {
...
   position: relative;
   overflow: hidden;
+  width: 400px;
+  height: 100px;
+  z-index: 10;
 }
 
 .donate-banner:hover {
   transform: scale(1.05);
+  animation: .5s background-fill forwards ease-out;
```
Alecaddd, line 308: this is an interesting solution but it doesn't give use full flexibility since we can't transition the background, so on mouse blur the effect is very jarring.
I'd suggest to use a pseudo element in the button to create the tilted red shape and then make that grow with an horizontal scale so we can have a smooth transition back and forth.

media/css/eoy.css
```diff
@@ -294,68 +475,160 @@ main {
...
+
+.donate-banner:hover #hover-hearts > #hover-heart-2 {
+  animation: 2s heart-fade forwards ease-in-out;
+  animation-delay: 0.3s;
+}
+
+.donate-banner:hover #hover-hearts > #hover-heart-3 {
+  animation: 2s heart-fade forwards ease-in-out;
+  animation-delay: 0.5s;
+}
```
Alecaddd, line 537: cute animation, I like it.
Let's make it a bit faster with each animation at 1s speed, and reduce the delay to 0.1s for heart-1 and 0.2s for heart-3, heart-2 doesn't need any delay.

Alecaddd (changes requested):
Great improvements, this is almost ready to go.
Make sure to do some quick x-browser testing since this page will be opened not just in Firefox, so at least guarantee parity with Chrome, Edge, and Safari.

media/css/eoy.css
```diff
@@ -59,18 +59,144 @@
 }
 
 :root {
-  --bg: #f9f9fa;
-  --text: #333;
+  --color-red-10: #FEE2E2;
+  --color-red-20: #FECACA;
+  --color-red-30: #FCA5A5;
```
Alecaddd, line 64: Do we need all these colors?

Alecaddd (approved):
It looks good, thanks

Alecaddd:
@Sancus tuis is good to go and it brings the EOY appeal to a production ready state

PR #340, opened 2022-11-24
Calendar Data Generation

### LINKED ISSUE #298: Automate Calendar Generation
We currently crowdsource calendar .ics files. This is error prone, and wastes time both ensuring correctness and reviewing update requests.

Considering there are now relatively high quality data sources like https://github.com/dr-prodigy/python-holidays or possibly https://calendarific.com/ there seems to be no good reason to continue with the manual, crowdsourced calendars.

MelissaAutumn:
Resolves #298

The future of automated holiday calendars is here!

Very simply it queries a calendar provider (which is currently Calendarific.), normalizes the data into a generic format, optionally mixes in additional holidays that may be missing from the provider. (An example is stored in `calgen/mixins/`) and writes it to individual ics files that are mapped via calendars.json.

If you'd like to add or change the data source provider, simply extend the Calendar model, and implement the from_api function. (You will also need to adjust the querying logic.) Very simple, and ensures any mixins will continue to work. An unused alternative is provided as an example in `calgen/models/HolidayAPI.py`.

For Calendarific you'll need to supply the environment variable `CALENDARIFIC_API_KEY` with your free or otherwise api key. If you're on a paid plan you can also set the environment variable `CALENDARIFIC_IS_FREE_TIER=false` to remove the 1 second sleep between requests.

Normally you could build the calendars through the `build-site.py --buildcalendar` script, but due to current Python 2 / Python 3 clashes, you'll have to call `build_calendars()` function from `build_calendar.py` manually.


build_calendar.py
```diff
@@ -0,0 +1,118 @@
...
+                        case requests.status_codes.codes.unauthorized:
+                            sys.exit("Bad or malformed API key")
+                        case _:
+                            print("Error {}: {}".format(err.response.status_code, err.response.reason))
+                            continue
+                except CountryNotSupportedException as err:
+                    print("Country code {} is not a supported locale.".format(err.country))
+                    continue
+
+        calendar_name = '{}Holidays.ics'.format(country_name)
```
Sancus, line 103: I would remove spaces from country_name so it doesn't make some filenames with them and some without.

build-site.py
```diff
@@ -34,7 +45,7 @@
     # Prepare data and build main website.
     version = helper.thunderbird_desktop.latest_version('release')
     beta_version = helper.thunderbird_desktop.latest_version('beta')
-    caldata = helper.load_calendar_json('media/caldata/calendars.json')
+    caldata = helper.load_calendar_json('media/caldata/autogen/calendars.json')
```
Sancus, line 48: Check if this file exists, and if not load the old one, because otherwise this will break builds on check-in.

build-site.py
```diff
@@ -21,9 +24,17 @@
...
     languages = ['en-US']
+    calendar_locales = [{
+        'US': 'United States',
+    }]
 else:
     langmsg = 'in all languages.'
     languages = settings.PROD_LANGUAGES
+    calendar_locales = settings.CALENDAR_LOCALES
+
+#if args.buildcalendar:
```
Sancus, line 35: Seems like this needs to be uncommented for this patch to run :)

Also, I would actually like `--buildcalendar` to run by itself only, so put it in as an `elif`. The calendar build process will run on a long schedule(once a month or less) and we'll just check the files into the repository. Rebuilding with every website build would also massively increase build times.

build-site.py
```diff
@@ -1,4 +1,6 @@
 import argparse
+
+#import build_calendar
```
Sancus, line 3: and this

build_calendar.py
```diff
@@ -0,0 +1,118 @@
...
+            year = current_year + i
+
+            for calendar_type in [CalendarTypes.NATIONAL, CalendarTypes.OBSERVANCE]:
+                try:
+                    holidays = query_calendarific(api_key, locale, year, calendar_type.value)
+                    formatted_holidays = [Calendarific(holiday, year, calendar_type) for holiday in holidays]
+
+                    for holiday in formatted_holidays:
+                        ical.add_component(holiday.to_ics())
+                except requests.HTTPError as err:
```
Sancus, line 88: The way this works right now, if there's any error, it writes an empty calendar file. That's not good, because we can assume that(after first run) there will be usable calendars in the tree. So if there's a network issue or anything during the build process, it will overwrite good calendars with empty ones.

If we have any error or no API data, we should just skip writing that file.

We *could* implement retries but I don't think it's really worth it, as long as we're not blowing up good data with bad.

calgen/mixins/GlobalHolidays.py
```diff
@@ -0,0 +1,23 @@
...
+
+These are mixed in before we pull data from the api source, and should use rrules.
+'''
+
+MIXINS = [
+    # Sample Event
+    # Thunderbird's Birthday
+    ##Calendar({
+    ##    'unique_id': '1254fbcc-762f-46b6-a85d-aa985c7776dc',
+    ##    'name': "Thunderbird's Birthday",
```
Sancus, line 16: Nice :)

build_calendar.py
```diff
@@ -0,0 +1,118 @@
...
+                    print("Country code {} is not a supported locale.".format(err.country))
+                    continue
+
+        calendar_name = '{}Holidays.ics'.format(country_name)
+
+        calendar_metadata.append({
+            'country': country_name,
+            'filename': "autogen/{}".format(calendar_name),
+            'datespan': date_span,
+            'authors': 'Autogenerated'
```
Sancus, line 109: nit: This fits in the sentence better if it's `Autogenerator`.

build_calendar.py
```diff
@@ -0,0 +1,118 @@
...
+from calgen.models.Calendar import CalendarTypes
+from calgen.models.Calendarific import Calendarific
+
+class CountryNotSupportedException(Exception):
+    def __init__(self, country):
+        self.country = country
+
+
+def query_calendarific(api_key, country, year, calendar_type):
+    if country not in settings.CALENDAR_LOCALES:
```
Sancus, line 21: I might be confused by something here, but how can this be called with a country that isn't in `CALENDAR_LOCALES` in the first place? It looks like `build_calendars` is called with `CALENDAR_LOCALES` as the initial list.

build_calendar.py
```diff
@@ -0,0 +1,118 @@
...
+        calendar_name = '{}Holidays.ics'.format(country_name)
+
+        calendar_metadata.append({
+            'country': country_name,
+            'filename': "autogen/{}".format(calendar_name),
+            'datespan': date_span,
+            'authors': 'Autogenerated'
+        })
+
+        with open('media/caldata/autogen/{}'.format(calendar_name), 'wb') as fh:
```
Sancus, line 112: nit: `open` will error if the directory hasn't been created yet. Since we are going to check calendars into the repository though, the directory should always be there after this patch.

Sancus (changes requested):
Seems mostly great except for the writing empty files issue!

Sancus:
Oh, also, please document your functions and classes [with docstrings](https://peps.python.org/pep-0257/) similar [to `builder.py`](https://github.com/thundernest/thunderbird-website/blob/master/builder.py#L79) - this doesn't always get done but it'd be nice to keep the standard high in this repo.

build_calendar.py
MelissaAutumn: Switched it out to just sys.exit for now. This can be adjusted if the provider is a bit spotty.

MelissaAutumn:
There is some issues with running build_sites.py on Python 2 (notably the calgen imports.) Hence why it was commented out to begin with. I'll look into that part. But otherwise, should be good.

MelissaAutumn:
Rebased, and some added changes to include localization where available, and to include local holidays. I've also removed the "How to contribute calendars" section. Let me know if we want to keep that but in a limited capacity.

I haven't included the built calendar files, let me know if I should do that. We can then just create a ticket to re-generate them quarterly.

Here's what our current list of calendars looks like:
Screen Shot 2023-06-07 at 13 59 50-fullpage

Sancus:
This PR needs rebasing and also the `requests` library needs to be added to requirements.

PR #341, opened 2022-11-25
RSS Feed for Release Notes

### LINKED ISSUE #264: RSS Feed for Release Notes page?
Any chance of creating a RSS feed for [release notes page](https://www.thunderbird.net/en-US/thunderbird/releases/) on thunderbird.net?

MelissaAutumn:
Resolves #264 

Includes rss feed generation into the build process. During the building of release notes, I gather up all the necessary details, and build the feed right after. This includes the full release notes. Tested and working with Thunderbird's RSS reader. 

I included beta releases by default. 

There's no new localization strings as the release notes are only in `en-US`.

In some brief research I found two rss generator projects
- https://github.com/lkiesow/python-feedgen
- https://github.com/getpelican/feedgenerator

I ended going with python-feedgen, as I had encoding issues with feedgenerator.

Example photo of the rss feed in Thunderbird.
image


MelissaAutumn:
I've updated the PR to switch from feedgenerator to python-feedgen due to encoding issues. I've tested this in Thunderbird's RSS reader, and it's working great. Did a spot check with the release note pages to ensure moving the content to a separate template does not harm the design. Looks fine there too.

builder.py
```diff
@@ -237,6 +243,78 @@ def build_notes(self):
...
+                continue
+
+            link = "{}/thunderbird/{}/releasenotes/".format(settings.CANONICAL_URL, version)
+
+            # Mix in our notes for the template
+            self._env.globals.update(**note)
+            # Hint to our svg helper function to skip svgs for this render
+            self._env.globals.update({'rss_build': True})
+            content = content_template.render({'version_number': version, 'link': link})
+            description = "For more on all the new features in Thunderbird {0}, see <a href=\"{1}\">What's New in Thunderbird {0}</a>".format(version, link)
```
jfx2006, line 297: This is kind of odd... will render as "For more on all the new features in Thunderbird 106.0beta, see"...

The issue is, not every version has new features. so this really isn't going to work.

jfx2006:
Can you send me a copy of the generated XML? The screenshot is nice, but there's things off-screen I'd like to see.

MelissaAutumn:
@jfx2006 Here you go. Let me know if you spot any issues. 
[feed.zip](https://github.com/thundernest/thunderbird-website/files/10118200/feed.zip)

builder.py
MelissaAutumn: Good point. I'll swap this out (and the footer of the template) for just the intro text then.

jfx2006:
image

- The title of every article includes "Release Notes", and the feed name also does. I think we can remove from article title/subject
- Website URL (under Subject) is wrong - last path component should be "releasenotes"
- Website URL also has wrong (102.5.0) version vs the Subject
- The bottom line "For more on new features..." - mentioned above that with the version like that it makes no sense and seeing it show up at the bottom of the article like that it looks out of place. I think it should just be removed 

Other comments:
Now that I have a full RSS feed file, I'm questioning the wisdom of including all of the notes in the feed itself. The file is 2.5MB. We don't need readers all over the place pulling that every 5 minutes. It's a lot of bandwidth.
One option may be to just publish the last 5 stable and last 5 beta releases in the feed.

For betas, Beta 1 will never have text in groups[0]. Some placeholder text may be needed to look consistent.

That whole description tag using the text header from the YAML doesn't really work out.. Might need to do something with the YAML files. I think we want everything up to where it talks about system requirements.

If we include the <content> tag, please use a separate template than the one used for the website. It makes no sense to include all of that extra html when some <ul> lists would work better. This will help with the overall size too.

Why are we publishing in RSS? Does it make more sense to use Atom? It seems to be the more advanced format?

Sancus:
> Now that I have a full RSS feed file, I'm questioning the wisdom of including all of the notes in the feed itself. The file is 2.5MB. We don't need readers all over the place pulling that every 5 minutes. It's a lot of bandwidth.

We don't pay for bandwidth so this isn't a huge concern, but there are a ton of ways to prevent RSS readers from downloading things every 5 minutes. We can set a TTL, use skiphours/skipdays, etc.

That said, I believe Atom supports some sort of pagination which might help as well.

> Why are we publishing in RSS? Does it make more sense to use Atom? It seems to be the more advanced format?

Yeah, we should use Atom. I forgot that it's treated as a different format, because all of this stuff that nobody uses definitely needed to be more obtuse than it already is lol.

MelissaAutumn:
Thanks for the feedback folks! I haven't personally used rss/atom since google reader died. :sweat_smile: 

I noticed the beta release notes only shows the initial release date, and not the updated date. Is this something we should change, or add an update time? I could do some fancy parsing in the `text` section, but I feel like that may not be sustainable. We can also ignore it / remove beta notes entirely if it's a big change. 

I just went with the last 10 releases, and last 5 betas. Right now that gives it a good mix of items in the feed, and the file is now like 150kb. 

I've pushed the updated changes, and I will revert the release-notes changes since I no longer need to touch that file. (So that's pending, everything else should be good.) 

Here's some updated examples:
image
Feed xml: [atom.zip](https://github.com/thundernest/thunderbird-website/files/10126767/atom.zip)

MelissaAutumn:
Whoops, noticed links were still incorrect. Here's the fixed build: [atom.zip](https://github.com/thundernest/thunderbird-website/files/10126893/atom.zip)

builder.py
```diff
@@ -306,26 +307,27 @@ def sort_version_by_release_date(i):
...
             self._env.globals.update(**note)
             content = content_template.render({'version_number': version, 'link': link})
 
             entry = feed.add_entry()
+            entry.id(link)
             entry.title(title)
             entry.link({'href': link})
-            entry.description(description=content)
-            entry.content(content, type="CDATA")
+            entry.content(content, type='html')
```
jfx2006, line 318: Maybe type="xhtml" here? Then it won't have to do all of the funny escaping. The template will need to be valid xhtml probably.

Either way, there's a lot of empty space after rendering the template. Possibly run the content through something like https://pypi.org/project/htmlmin/, or use something like https://pypi.org/project/pyquery/ to build programatically? Neither option is great since they have dependencies...

jfx2006:
> I noticed the beta release notes only shows the initial release date, and not the updated date. Is this something we should change, or add an update time? I could do some fancy parsing in the `text` section, but I feel like that may not be sustainable. We can also ignore it / remove beta notes entirely if it's a big change.

Right.. I was thinking about how to deal with this. We could extend/replace "groups" with release date info for beta >1. The other option, since product-details data is around somewhere during the website build, you can grab release dates  for those betas out of there somehow.

Since betas >1 are "updates" to existing feeds, there's an "updated" value that should be set for those.

The rest...

Subject lines... "Thunderbird 102.5.1" for stable releases and "Thunderbird Beta 108.0" for betas?

"Changes"  &  "Fixes" headers - - please make these past tense. (The website should be changed as well probably)

The header info coming from "text" in the YAML -- I think we don't need the requirements included in the feed for each release.  I think it can be pulled out of the website version too and just link. I'll work on that.

website/_includes/release-notes-feed.html
```diff
@@ -0,0 +1,77 @@
+{% set is_beta = channel_name == "Beta" %}
+<p>{{ _('Check out the notes below for this version of Thunderbird. As always, you’re encouraged to <a href="%(feedback)s">tell us what you think</a>, or <a href="%(bugzilla)s">file a bug in Bugzilla</a>')|format(feedback='https://support.mozilla.org/questions/new/thunderbird', bugzilla='https://bugzilla.mozilla.org/') }}</p>
```
jfx2006, line 2: For the feed, lets remove this line. At the bottom after the notes, put the "As always... Bugzilla" links.

builder.py
MelissaAutumn: I went with htmlmin. It'd be nice to eventually minify everything, since it seems to be a common issue through the site. There are settings for jinja to strip white spaces, but I couldn't seem to get it working with just the feed. (or at all, actually :thinking:)

As for the xhtml, the python markdown library seems to be acting weird with missing quotes around the src attribute. I'm poking at it a bit more though. Although I don't think the funny escaping is a particular issue, unless you've encountered problems with it in a reader I haven't tested.

jfx2006: The escaping is a readability problem more than anything else. But if we are minifying too that argument doesn't hold up. :-)

jfx2006: The Jinja version on production is possibly too old for whitespace stripping?

MelissaAutumn: Locally (from requirements-dev) I've got 2.11.3 installed, and looks like it should be supported: https://jinja.palletsprojects.com/en/2.11.x/templates/#whitespace-control (lstrip_blocks, and trim_blocks). But alas, settings those in the Environment constructor does nothing to the feeds output. 

There's a few other issues I've encountered trying to set it to xhtml, mostly crypticly set error messages (despite the modified builds producing xhtml valid markup :nerd_face:).

MelissaAutumn:
Latest build: [atom.zip](https://github.com/thundernest/thunderbird-website/files/10135590/atom.zip)

No exciting changes aside from the minor text changes. Thanks for pointing out the product details, managed to get the specific beta's release date.

MelissaAutumn:
Two things that came to mind over the break: (Open to any answers.)
- If the feed content is still too messy, I'll gut the feed generator package, and just render it as a jinja template. (Similar to how WordPress handles their feed generation.)
- We probably need a feed icon on the release notes page, right?

MelissaAutumn:
I've squashed it, and fixed it up for Python3 in case we want this to go out anytime soon. 😄 

I didn't add a fancy rss icon, but it's in a meta tag for the releases page. 

Latest sample:
[atom.xml.zip](https://github.com/thundernest/thunderbird-website/files/11871588/atom.xml.zip)

Perkolator:
> I included beta releases by default.

I'm really late to the party, already forgot about my original issue I created 2 years ago.

Why include betas? Isn't this feed supposed to be a feed for the [Thunderbird Release Notes](https://www.thunderbird.net/en-US/thunderbird/releases/) webpage? That doesn't include betas. I wouldn't want to see betas in the feed.

Perkolator:
> * We probably need a feed icon on the release notes page, right?

Yes, definitely. I never have understood webpages that have feeds but don't indicate that it/them are available and the user needs to check the source code of the page to see if there's feed(s) available.

MelissaAutumn:
Yea good point. I've added a basic link to the releases page, I've kept the setting as we may generate a beta or mixed release atom feed in the future in addition to the ESR feed.

Perkolator:
No one has time to review?

Perkolator:
> The frustration is relatable, though not helpful.

I don't man, sometimes a little kick could be quite stimulating. ;) And it's not like I'm pestering them all the time.. look at the dates.

Perkolator:
Is there something still blocking this?

@jfx2006
@Sancus

MelissaAutumn:
Hey everyone! 

With the launch of the new website I'm currently re-working this to remove external dependencies. There will be a new PR up soon and should functionally be identical. 

Thank you for your patience!

PR #344, opened 2022-12-05
Responsive and l10n fixes for donate button

Alecaddd:
Some small improvements to the donate button to let it play nice with long translated strings. Example in Italian where the text doesn't wrap properly and during hover on small screens it doesn't look readable.

Before
Screen Shot 2022-12-04 at 4 50 52 PM

After
Screen Shot 2022-12-04 at 5 11 58 PM

### DIFF
--- a/media/css/eoy.css
+++ b/media/css/eoy.css
@@ -180,7 +180,6 @@ main {
 
 .donate-banner {
   --accent-text: var(--color-red-80);
-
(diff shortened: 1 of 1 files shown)


micahilbery (approved):
I tried a few different languages and it looks ok to me. :+1:

Sancus:
@micahilbery Thanks!

PR #347, opened 2022-12-06
Outdated page

### LINKED ISSUE #170: Create general ugly red warning start page for old versions
Right now we have the earlybird warning page: https://start.thunderbird.net/en-US/earlybird/

It's not localized, and we should probably change the copy to not talk about Earlybird:

"This version is no longer a supported release, and does not receive security updates!"
"We **strongly recommend** downloading the latest stable version of Thunderbird."

After localization, we should redirect start pages to it. I suggest everything below Thunderbird 52(So 45, 38, and everything below that would be the major versions affected). This is roughly 4% of our users.

MelissaAutumn:
Resolves #170 

This updates the page to be on the newest start template, and grabs the dates/versions programmatically based off the new `settings.LAST_SAFE_VERSION` variable. If we could grab the last safe version from product details that would be rad, but this should make it quite simple to update. 

We're still missing the support article, but I can also just make the line about the support article more generic. 

Screenshots:
Thunderbird v10.0.1
v10 0 1
Thunderbird v31.8.0
v31 8 0
Thunderbird v78.9.1
v78 9 1

During testing I noticed we use Let's Encrypt which has deprecated [an older CA](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/) a while ago. On an older system that hasn't been updated with the new root certificate this will cause a white screen (if it's in the message body section) or an certificate error page. Probably not an issue, because it's unlikely that someone would be this outdated (except for my virtual machine), but something to keep in mind.
v10 0 1-cert

### DIFF
--- a/helper.py
+++ b/helper.py
@@ -5,6 +5,8 @@ import jinja2
 import json
 import markdown
 import re
+
+import product_details
 import settings
 import sys
 import translate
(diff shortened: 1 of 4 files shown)


MelissaAutumn:
I've updated the design to fix contrast issues and make it redder. 
Screen Shot 2023-11-24 at 10 47 07-fullpage

MelissaAutumn:
Updated copy:
image

MelissaAutumn:
I'll get this merged soon after re-design. We've removed font-awesome from our main directories so I'll have to adjust the icon but nothing too big.

PR #351, opened 2022-12-06
Allow localized beta downloads

### LINKED ISSUE #350: Localize beta links using release locales
Lets have beta use the locale information from the release primary builds for now so we can localize these URLs.

I'm not totally sure which code needs to change for this, probably something in `helper.download_thunderbird` or somewhere in the `product_details.py` stuff.

MelissaAutumn:
Resolves #350 

I'm using the `all` key here, which only has beta and daily information. But I can use en-US if that works better.  I'm filtering the `all` key for beta builds only. So daily will still be en-US.

Tested it out locally with all the languages that have localized releases, and works great. (And now I have about 65 builds in my downloads folder.)

### DIFF
--- a/product_details.py
+++ b/product_details.py
@@ -20,6 +20,28 @@ def load_json(path):
     with open(path, 'r') as f:
         return json.load(f)
 
+def load_all_builds(path):
+    """ Loads the all_builds data, and mixes in the latest beta information with all release locales. """
+    all_builds = load_json(path)
+
+    if 'all' not in all_builds:
+        return all_builds
+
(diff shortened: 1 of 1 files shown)


Sancus:
This looks good to me, @jfx2006 what is the 'all' locale and can we rely on it staying there? I don't recall what it's for.

Sancus:
We should use `en-us` for this.

Sancus (changes requested):
Looks good, but we should use `en-us` for this.

PR #356, opened 2022-12-15
Fundraiseup integration

MelissaAutumn:
Some initial support for Fundraise Up. 

I've added a super simple AB test module to pick 0 or 1, and we conditionally use Fundraise up or the old flow depending on that choice. This choice isn't saved, so you can keep refreshing. Currently it will also spit the choice out in the console log. (0 = Fundraise Up, 1 = old flow.) We can add complexity as needed, I figured for our use-case this was probably a good plan. I guess we'll still need a way to track donations between platforms, but maybe that could just be a donation platform report instead?

Fundraise Up is only hooked up on the front-page download button, but I included the api js snippet on any page that uses that donation modal. (Which currently seems to be all of them.) So let me know if I should just go ahead and swap all the donate links to the bucket determined flow.

It is in live mode, I intended to leave it on test mode but it adds an annoying bottom bar to every page, grr. 

I've tested the location.href to download link on Firefox, Safari, Edge (VM), IE11 (VM). They all seem to work, the page has other styling issues on IE11, but eh. 

Oh yeah, I guess we can remove the braintree stuff right?


assets/js/common/mozilla-utils.js
```diff
@@ -40,60 +40,72 @@ if (typeof Mozilla === 'undefined') {
...
         $('#submit-button').hide();
-        $('.download-link').each(function() {
+        $('.download-link').each(function () {
             var $el = $(this);
-            $el.click(function(e) {
-                e.preventDefault();
-                window.Mozilla.Donation.DisplayAmountForm();
-                // Utils.triggerIEDownload($el.data('direct-link'));
+            $el.click(function (e) {
+                if (window.ABTest.IsInFundraiseUpBucket()) {
```
Sancus, line 48: It feels like a bunch of this stuff should be in a function. Also, lets leave the AB test off for now because it seems like it'll be a bit annoying while we're just prototyping the new flow itself.

Sancus (changes requested):

assets/js/common/mozilla-utils.js
```diff
@@ -20,7 +20,7 @@ if (typeof Mozilla === 'undefined') {
      * @param {userAgent} optional UA string for testing purposes.
      */
 
-    Utils.triggerIEDownload = function(link, userAgent) {
+    Utils.triggerIEDownload = function (link, userAgent) {
```
Sancus, line 23: Spaces before function parens is inconsistent with all the other JS in this repository, so all these lines shouldn't be here. Making style changes in adjacent code like this also just adds a lot of noise to PRs.

assets/js/common/donations.js
```diff
@@ -3,133 +3,82 @@ if (typeof Mozilla === 'undefined') {
...
-        });
+        // Ensure we actually have the javascript loaded, so we can hook up our events.
+        if (window.FundraiseUp) {
+            const fundraiseUp = window.FundraiseUp;
+            // Close our modal on open
+            fundraiseUp.on('checkoutOpen', function () {
+                Donation.CloseForm();
+            });
+            // Start the download on completion
+            fundraiseUp.on('donationComplete', function () {
```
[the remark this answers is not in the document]
MelissaAutumn: Shall I change the text to Download & Donate as well? 😄

PR #367, opened 2023-01-16
FRU: Fill in window._lang

### LINKED ISSUE #361: Fill in window._lang for FRU
So, in order to make sure the donation modals are the same language as the website, we need to pass the l10n info to FRU.

There is some documentation here: Make website l10n consistent with FRU - (probably use window._lang) https://fundraiseup.com/docs/localization-settings/

Also we should make sure that they fall back correctly to English if we pass an unsupported language.

MelissaAutumn:
Resolves #361 

Supported languages: (Slightly vary based on what our account supports)
https://fundraiseup.com/docs/languages/

Questions:
- Should we include 'nn-NO' as 'nb'? (They're both Norwegian, but I'm sure there's some regional differences.) 
- Should we include Spain Spanish under `es`? (It's labelled as Latin American Spanish, so I'm guessing no 🤔)

Pytest can be run via `python -m pytest tests/` in the base directory. Once the py3/calgen stuff gets merged in, I'll clean that stuff up and add a readme message about it.


Sancus:
> Resolves #361
> 
> Supported languages: (Slightly vary based on what our account supports) https://fundraiseup.com/docs/languages/
> 
> Questions:
> 
>     * Should we include 'nn-NO' as 'nb'? (They're both Norwegian, but I'm sure there's some regional differences.)
> 
>     * Should we include Spain Spanish under `es`? (It's labelled as Latin American Spanish, so I'm guessing no 🤔)
> 

Latin Spanish is definitely mutually intelligible with Spain Spanish so I would say yes. Someone from Spain is going to have a much easier time with that than they would have with English.

I think nn-NO and nn-NB are also mutually intelligible but not totally certain. I would just go for it and if anybody complains we can change it.

PR #368, opened 2023-01-16
FRU: replace all donation links with fru

### LINKED ISSUE #366: Update all donation links on the website to use FRU

MelissaAutumn:
Resolves #366 

Also fixes a bug introduced in https://github.com/thundernest/thunderbird-website/pull/356 (oops, my b) that seemingly broke the downloads for beta/daily.

This method does technically reload the page, but that's technically what the custom FRU button element does. So it's a little jarring on buttons further down the page.


MelissaAutumn:
For https://github.com/thundernest/thunderbird-website/issues/362
I've brought UTM tag support that should match the legacy flow's UTM tags, and also adjusted the module the donation form popping up code lives.

PR #369, opened 2023-01-17
FRU: UTM Tags for FRU

### LINKED ISSUE #362: UTM tags for FRU donations
We need to figure out UTM tags for FRU: https://fundraiseup.com/support/utm-parameters/

MelissaAutumn:
Resolves #362 

We're using the utm campaign: `donation_flow_2023` and adjusting the utm medium to be `fru` (for FundraiseUp) and `give` (for the legacy give.thunderbird.net flow). 

This includes the addition of FRU on the start site, and EOY. start page changes surprisingly looks okay in-app. This will need to be added to the test plan though.
image

I've also added a maybe fix for the jarring page reload to the top issue we're having.


assets/js/common/donations.js
```diff
@@ -133,6 +139,21 @@ if (typeof Mozilla === 'undefined') {
         }
     };
 
+    /**
+     * Checks and applies any position parameter to the window's scrollY
+     * This makes the silly page refresh on the donation modal look less jarring
+     */
+    Donation.CheckForPosition = function () {
```
Sancus, line 146: Hmm... I'm not sure this is actually working? In my testing on Firefox, it didn't seem to change the behaviour. I still get click download beta -> page reload showing top of page with modal -> after closing modal, page resets to bottom.

MelissaAutumn:
Turns out the button download issue was due to window.location. I've swapped it for a window.open(link, _self), which I tested on FF, Edge, Chromium, and Safari. Seems to work fine. 

The donate link solution is a little janky, but it more or less works in the cases I've tested.

MelissaAutumn:
Good point, I've reverted that change!

PR #372, opened 2023-01-23
Create a newsletter landing page, and consolidate the newsletter form.

### LINKED ISSUE #371: Newsletter Signup Page
We're going to need this as part of the FRU stuff.

I'm aware this was already done on the blog page, but it's really not useful there for the reasons I specified in https://github.com/thundernest/thunderblog/issues/34#issuecomment-1363582446

MelissaAutumn:
Resolves #371 

I've re-used the existing form, and added the "See previous newsletters" link I used on the WP side. I've also done some tweaks to ensure that the mobile view is as decent as possible. (This being a newsletter landing page and all.)

Some questions/concerns:
- Do we want the header and footer on this page, or is this more of a stand-alone page?

- I find the actual newsletter form, and the standard small-link anchour tags a bit too small for me. I feel like we could double the size without too much issue there, but it would break consistency with the rest of the site. (And the pre-footer.)

Screenshots:
Full width (1800w):
image

Half width (900w):
image

Iphone 12 mini:
image


Sancus:
@micahilbery Hey Micah, can you take a look at this and offer any suggestions you might have design-wise?

I think we should definitely remove the pre-footer at least, and maybe the header as well. Also, the signup form should go up top of the text, be centered, and the button definitely needs to be larger. The call to action here is to sign up for the newsletter, we don't want to detract or distract from that and we want people to gravitate to doing that immediately.

@KillYourFM Any comments?

KillYourFM:
The call to action here is to sign up for the newsletter, we don't want to detract or distract from that and we want people to gravitate to doing that immediately.
> 
> @KillYourFM Any comments?

The mobile version is perfect! 
On the desktop pages, I would suggest removing the Download and Contribute links. This page should only have a single call to action to maximize signups. It's a nitpick though, and it's just my opinion.

MelissaAutumn:
Yeah I'll be knocking off the header, and pre-footer (Download and Give sections.), and making the entry field larger, as it's so tiny right now.

MelissaAutumn:
So I've played around with a few different styles, and I've attached the "good" looking ones below. I think the page could definitely benefit from having more content. 

Maybe we could add a "Check us out on these social media platforms" (but worded better...) with larger icons for each site. Additionally we could hide the social media icons in the footer for this page only. 

I've checked out a few other orgs newsletter CTA / landing pages, and they all tend to be very small, and devoid of content. 🤔 

Bigger, but more or less the same:
image
Bigger alternate styling:
image

Any ideas folks?

KillYourFM:
> So I've played around with a few different styles, and I've attached the "good" looking ones below. I think the page could definitely benefit from having more content.
> 
> Maybe we could add a "Check us out on these social media platforms" (but worded better...) with larger icons for each site. Additionally we could hide the social media icons in the footer for this page only.
> 
> I've checked out a few other orgs newsletter CTA / landing pages, and they all tend to be very small, and devoid of content. 🤔
> 
> Bigger alternate styling: image
> 
> Any ideas/suggestions/comments/rubber stamps folks?

This one ("bigger, alternate styling") really pops. 

Some suggestions:

1. Perhaps center the "Join Our Newsletter" heading so it center-aligns with the rest of the content. 
2. Let's keep the social icons, but move up the existing row of social icons to below the email signup box. Make them larger as you said, and extend the blue page background beneath those. Ensure that if clicked, they open in a new tab. Let's not do a separate heading or callout text for the social destinations, though. I think it would add too much noise to the intent of the page. 
3. Then start the grey background / footer area where the Thunderbird logo is now.

MelissaAutumn:
@KillYourFM (and maybe @micahilbery) How's something like..

Wide shot:
image
Not Wide shot:
image
Iphone 11 miniii:
image
image
Hover interaction: (Standard, but just as refresher)
image


I had to drop the ! off the title as, while it is centred the ! added to much visual weight and made it look not centred. 🤔 

I've been playing with a "light" theme version of the social media buttons, since the standard dark appearance against the bright blue wasn't great. But something feels off about them..

But the social media links seem to balance it out quite nicely, so it's a good step.

Alecaddd:
The white box on white input field with light grey text is not great for a11y.
Colors should be tweaked a bit. I'll suggest some edits in a bit

website/newsletter/index.html
```diff
@@ -0,0 +1,68 @@
...
+
+{% block header_content %}
+  <section class="text-center text-white flex justify-between pl-4 pr-4 sm:pl-8 sm:pr-8 md:pl-16 md:pr-16 max-w-7xl mx-auto pt-12 mb-20">
+    <aside class="flex-col items-start">
+      <h1 class="font-4xl md:font-hero font-thin mt-0 mb-8">{{ self.page_title() }}</h1>
+      <p class="font-lg md:font-xl tracking-wide leading-relaxed max-w-4xl lg:mb-16 mb-8 mt-0 p-links mx-auto">
+          {{ _('See exciting previews of upcoming features, read the latest desktop Thunderbird and mobile Thunderbird news, and enjoy personalized tips to protect your privacy and maximize your productivity! From the Thunderbird team directly to your inbox.') }}
+      </p>
+
+      <div class="flex flex-col items-center mb-20 mx-auto max-w-xl bg-grey p-4 rounded shadow-lg text-black-light no-underline md:font-md">
```
Alecaddd, line 20: I would remove the box shadow and grey background from this.

website/_includes/newsletter-form.html
```diff
@@ -0,0 +1,13 @@
+<!-- Begin Mailchimp Signup Form -->
+<form action="https://thunderbird.us12.list-manage.com/subscribe?u=f8051cc8637cf3ff79661f382&id=56428f2efc"
+class="shadow flex items-center rounded-md self-start p-1"
```
Alecaddd, line 3: for this I would use a grey-lighter background and let the child items stretch its full height.
Also maybe a 6px gap.

assets/less/components/links.less
```diff
@@ -73,6 +73,29 @@ a {
   }
 }
 
+.social-link-large {
+  &:extend(.social-link, .font-regular, .bg-grey, .text-blue, .font-semibold, .shadow-md, .mb-4);
```
Alecaddd, line 77: let's use a `text-blue-dark` for these, as well as the background color on rollover in order to guarantee AA readability.
The firefox inspector will also complaint if the AA is below 4points.

website/newsletter/index.html
```diff
@@ -0,0 +1,68 @@
...
+        <div class="w-full">
+        {% include '_includes/newsletter-form.html' %}
+        </div>
+        <p class="font-sm font-semibold self-start">
+          <a class="small-link font-regular text-blue" href="https://us12.campaign-archive.com/home/?u=f8051cc8637cf3ff79661f382&id=56428f2efc" title="View previous newsletters">{{ _('View previous newsletters') }}</a>
```
Alecaddd, line 25: make this `text-white`

assets/less/components/buttons.less
```diff
@@ -117,7 +117,7 @@ button {
 }
 
 .btn-newsletter {
-  &:extend(.rounded-xs, .border-none, .bg-primary-button, .text-white, .p-2, .pl-3, .pr-3, .uppercase, .font-semibold, .appearance-none);
+  &:extend(.w-40, .h-8, .rounded-xs, .border-none, .bg-primary-button, .text-white, .p-2, .pl-3, .pr-3, .uppercase, .font-semibold, .appearance-none);
```
Alecaddd, line 120: Remove the fixed height so we let the button stretch the full height.

website/newsletter/index.html
```diff
@@ -0,0 +1,68 @@
...
+{% block header_content %}
+  <section class="text-center text-white flex justify-between pl-4 pr-4 sm:pl-8 sm:pr-8 md:pl-16 md:pr-16 max-w-7xl mx-auto pt-12 mb-20">
+    <aside class="flex-col items-start">
+      <h1 class="font-4xl md:font-hero font-thin mt-0 mb-8">{{ self.page_title() }}</h1>
+      <p class="font-lg md:font-xl tracking-wide leading-relaxed max-w-4xl lg:mb-16 mb-8 mt-0 p-links mx-auto">
+          {{ _('See exciting previews of upcoming features, read the latest desktop Thunderbird and mobile Thunderbird news, and enjoy personalized tips to protect your privacy and maximize your productivity! From the Thunderbird team directly to your inbox.') }}
+      </p>
+
+      <div class="flex flex-col items-center mb-20 mx-auto max-w-xl bg-grey p-4 rounded shadow-lg text-black-light no-underline md:font-md">
+        <div class="w-full">
```
Alecaddd, line 21: Add a `bg-grey-light` and a 4px border radius to this.

Then the input field and button should have a 2px border radius.

Alecaddd (changes requested):
I left some comments and suggestions to improve the style and accessibility.
The final result should look like this. Some changes should also affect the look of the newsletter form used in the home page.
Screenshot_20230127_032800

MelissaAutumn:
Not wide:
image

Wiiiide:
image

Mobile-ish: (Accurate to my phone)
image

Adjusted Hover:
image

I've done some additional mobile adjustments. How's the icon background colour on the social icons? It should probably match, huh. 

...I need to fix the white gap at the bottom

micahilbery (approved):
Looks great!

PR #374, opened 2023-01-24
Features/370 donation landing page

### LINKED ISSUE #370: Landing page for donations
Now that we will be moving to FRU, https://give.thunderbird.net will be retired at some point.

We are still going to need a landing page for donations, for better SEO and to have a place for [the FAQ](https://give.thunderbird.net/en-US/faq/) and [Ways to Give](https://give.thunderbird.net/en-US/ways-to-give/) content.

In order to reduce the number of steps, we should embed the donation form in this one rather than using the modal. Hopefully that's not too much extra work.

First pass could use a two column layout similar to https://give.thunderbird.net/

https://www.thunderbird.net/donate/ seems like a good URL for this.

MelissaAutumn:
Resolves #370 

Adds a donation landing page, which is available at `/donate`.

Notes:
I had a bit more content to play with, so it looks a little less empty. It's all on one page, and I nabbed the "Features" page's table of contents for this. Overall I think it flows quite nicely, however I think we could probably find a way to spruce up the FRU element.

~~The faq entry block of text kind of scares me, I could probably move that to a .py file.~~
I've re-worked the faq entries into a helper function, it's a lot easier to digest and modify now. 

For other ways to give, I tried to keep text localizable based on the source: https://give.thunderbird.net/fr/ways-to-give/ The lack of a gettext call for financial information is intentional.

Reference Content:
https://give.thunderbird.net/en-CA/faq/
https://give.thunderbird.net/en-US/ways-to-give/

Screenshots (Wide):
image
image
image
image

Screenshots (not wide):
image
image
image
image


MelissaAutumn:
Fixed string extractions with the FAQ.

Also ran into some annoying Mac issues with l10_tools (FreeBSD tools vs Gnu tools), so I at least documented those. I'll make a note to re-work some of those scripts.

micahilbery (approved):
As far as the design goes LGTM! Great work!

micahilbery:
As I said earlier it looks good to me. However, @Alecaddd might want to have a look as well.

website/donate/index.html
```diff
@@ -0,0 +1,81 @@
...
+        </h3>
+        <p class="font-lg tracking-wider leading-loose mb-6">
+          {{ _('Our mission is to provide an open source email app that is secure, private and free for users around the world. Your donations support this work. Will you give today?') }}
+        </p>
+        <p class="font-md tracking-wider leading-loose mb-6">
+          {{ _('Contributions are not tax-deductible as charitable donations.') }}
+        </p>
+      </aside>
+      <aside class="flex-1 max-w-xl text-center mx-auto w-full">
+          <a href="#XVFNMBAK" style="display: none"></a>
```
Alecaddd, line 50: nit: wrong indentation
```suggestion
        <a href="#XVFNMBAK" style="display: none"></a>
```

Alecaddd (approved):
Just a quick fly by review since my build on Windows seems to be broken, so I can't see this in detail.
I would suggest to create a bit more space in between questions in the FAQ section, and separate them with the same style we currently use for version: https://www.thunderbird.net/en-US/thunderbird/releases/

Is that inset round shadow at the bottom of your screenshots something generate by the screenshots or it's there in the page?

I would make the + and - icons like a `.header-section span`.

These are just styling nits and can be done later as I don't want to be a blocker on this page.

PR #394, opened 2023-02-14
Features/391 post download thank you

### LINKED ISSUE #391: post-download thank you/installer page
We should have a page similar to Firefox's that tells people that their download was successful and they should go install the thing.

https://www.mozilla.org/en-US/firefox/download/thanks/

We should probably just change this page accordingly since it doesn't really have much of a purpose with the FRU workflow anymore https://www.thunderbird.net/en-US/download/

MelissaAutumn:
Resolves #391 

I have a feeling this might go through a few design revisions :sweat_smile: 

Also I want to casually tag @KillYourFM for a copy check. 

Screenshots:
image

No Javascript:
image

~Adjusted the "Download didn't work?" message to a banner, I stole the banner yellow from Thunderbird's remote content block message. I don't expect no-javascript folks to actually get to this page, but just in case I've made it look "acceptable."~

Didn't notice that we stick `no-js`/`js` on the html, that came in handy. Thanks @micahilbery for the design direction! 

I merged in the new download page changes, and uhh. Well I guess I could bring in the download channel bit from pre-footer, but again I don't think many no-javascript folks will get to this page.


Sancus:
This page is going to be pulling double duty basically, as both a page that can trigger a download and as the final step in the (already-completed) download button process.

So we should just have a GET boolean like `?downloaded=false/true` with a default value of `false`. The download button process can then link to this page with `?downloaded=true`. This will still give people a confirmation that their download completed and a pointer to what to do next, plus an easy way to try again in the odd case it somehow failed.

assets/js/common/autodownload.js
```diff
@@ -6,9 +6,15 @@
...
+    if ($('body').attr('id') === 'thunderbird-download') {
+        // IE11 safe check for determining if we've already downloaded Thunderbird
+        var lowercaseQuery = window.location.search.toLowerCase();
+        var isDownloaded = /([?|&]+downloaded=true)/.test(lowercaseQuery) === true;
+        if (isDownloaded) {
+            return;
+        }
+
         var isIELT9 = window.Mozilla.Client.platform === 'windows' && /MSIE\s[1-8]\./.test(navigator.userAgent);
-        var $directDownloadLink = $('#direct-download-link');
```
MelissaAutumn, line 11: Unused variable

assets/js/common/autodownload.js
```diff
@@ -6,9 +6,15 @@
     'use strict';
 
     // Only do this on the autodownload page.
-    if ($('body').attr('id') == 'thunderbird-download') {
+    if ($('body').attr('id') === 'thunderbird-download') {
+        // IE11 safe check for determining if we've already downloaded Thunderbird
+        var lowercaseQuery = window.location.search.toLowerCase();
+        var isDownloaded = /([?|&]+downloaded=true)/.test(lowercaseQuery) === true;
```
MelissaAutumn, line 12: IE11 doesn't support SearchParams API, so a little regex here should ensure it doesn't break it.

MelissaAutumn:
Adjusted "Download didn't start" and font spacing.

MelissaAutumn:
Updated the retry download banner. Looking good now.

micahilbery (approved):
Great work! Design LGTM :+1:

micahilbery:
There could be a few refinements around the no js download dialog but I would need to spend some time thinking about it and I don't want to be a blocker.

MelissaAutumn:
Ah right. @Sancus either this will need to get merged for the redirect flow that went onto stage, or I can also split the downloaded=true/false variable from this into its own PR.

PR #409, opened 2023-02-28
Remove Donation Flow 2023 AB Test (Fixes #408)

### LINKED ISSUE #408: Remove A/B Test

MelissaAutumn:
- Clean up some now unneeded code
 - Comment out the rest that may be useful later
 - Enable redirect to downloads page after download has begun
 - Move FRU specific donation url generation to our build process (Python)
 - Removed JavaScript from start page for now

### DIFF
--- a/helper.py
+++ b/helper.py
@@ -318,19 +318,48 @@ def thunderbird_url(page, channel='None'):
 
 
 @jinja2.contextfunction
-def donate_url(ctx, content='', source='thunderbird.net', medium='give', campaign='donation_flow_2023', download=False):
-    # If this link is from a download button, donate.mozilla.org has thank you text.
-    download_string = ''
-    if download:
-        download_string = '&tbdownload=true'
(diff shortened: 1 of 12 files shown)


helper.py
```diff
@@ -318,19 +318,48 @@ def thunderbird_url(page, channel='None'):
...
+        'utm_medium': medium,
+        'utm_campaign': campaign,
+        'form': form,
+        'downloaded': download,
+        'download_channel': download_channel
+    }
+
+    query_list = []
+
+    for item in query.items():
```
Sancus, line 347: dictionary -> url parameters is a common operation and can be done with `urllib.urlencode` instead. `urllib.parse.urlencode` in Python 3.

MelissaAutumn: Thanks for the catch! Now if only there was a dict filter for None values 😅

Sancus: Yeah, you need to remove the `None`s, but a dict comprehension is a good way to do this:
`filtered_query = {k: v for k, v in query.items() if v is not None}`

PR #420, opened 2023-04-03
Replace references of Lightning Calendar to just Calendar

MelissaAutumn:
Replace references of the old Lightning Calendar to just Calendar, and correct some links so they point to the right place.

I thought about using Thunderbird Calendar, but Calendar seems to be the way it's referenced elsewhere.

### DIFF
--- a/website/calendar/index.html
+++ b/website/calendar/index.html
@@ -4,7 +4,7 @@
 
 {% extends "includes/base-resp.html" %}
 
-{% block page_title %}{{ _('Lightning Calendar') }}{% endblock %}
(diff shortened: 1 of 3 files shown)


website/calendar/index.html
```diff
@@ -84,7 +84,7 @@ <h4 class="header-section mb-0">
       </aside>
 
       <aside class="flex flex-col w-1/2">
-        <a href="https://codetribute.mozilla.org/projects/calendar"
```
Sancus, line 87: Hmm it kinda sucks that they removed this. Oh well.

PR #422, opened 2023-04-03
Beta Appeal

MelissaAutumn:
Some initial work on the beta appeal

### DIFF
--- a/assets/js/common/beta-appeal.js
+++ b/assets/js/common/beta-appeal.js
@@ -14,14 +14,8 @@ if (typeof Mozilla === 'undefined') {
      * @constructor
      */
     BetaAppeal.Join = function (evt) {
-        if (evt.currentTarget.dataset.joined) {
-            return;
-        }
-
-        const joinBtn = evt.currentTarget;
-
-        // Don't run this again
-        joinBtn.setAttribute('data-joined', true);
+        document.getElementById('ba-step-1').classList.add('hidden');
+        document.getElementById('ba-step-2').classList.remove('hidden');
     }
 
     /**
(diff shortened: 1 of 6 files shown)


emitche:
I am a writer, so if you need any input on copy, feel free to reach out. This looks beautiful so far based on the screenshot shared.

Sancus:
> I am a writer, so if you need any input on copy, feel free to reach out. This looks beautiful so far based on the screenshot shared.

I shared the doc with you and other people who expressed interest, though anyone on the team can use the link to view/edit.

wsmwk:
This will need to be removed?  "When Thunderbird Supernova officially launches later this Summer, you will be switched back to the stable, non-Beta version automatically."

wsmwk:
The appeal is now for users to install beta in parallel (not be upgraded)  so https://docs.google.com/document/d/1XJ7rymJw4OiGLr-KvXnJ5I6AhtwjIASI3s9_3WL6J8s/edit#heading=h.x3pg6hodgnfc is obsolete and we must consider additional information for the beta appeal page.  

Over 40% (?) of users have pop accounts.  I bet most most of these pop users are not technically adept and will be very poor candidates for running two instances of Thunderbird - some of them will definitely shoot themselves in the foot. It is the reason we added cautionary language in https://www.thunderbird.net/en-US/download/beta/   

Perhaps the page should suggest that imap only users should apply.  OR tell pop users to consult the KB article.

Also I'm pretty sure in past meetings that 
a) I mentioned we really need language to users that we want their feedback, and where they do it  (Help menu may be sufficient)
b) @Sancus or someone said we would be giving the link to the beta KB article, which I don't see listed in the current appeal

Do we need to create a new issue for these items?  Or use #421 ?

Sancus:
I'm going to merge this as-is with the understanding that it's still WIP, just to make it easier so we have the page on stage and can reference it.

PR #443, opened 2023-05-24
Manually track the donation modal open goal (Fixes #441)

### LINKED ISSUE #441: Fix donation link click goal
The goal broke once we stopped linking to give.thunderbird.net.

MelissaAutumn:
The event "Donation link click" broke when we removed give.thunderbird.net and switched to FRU. While we do track an event per modal open, we should also track the link clicks themselves, as 'checkoutOpen' is automatically opened in some cases. (Like after the download page.)

### DIFF
--- a/assets/js/common/donations.js
+++ b/assets/js/common/donations.js
@@ -32,6 +32,15 @@ if (typeof Mozilla === 'undefined') {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


radishmouse (approved):
:+1: solid JavaScripting :)

PR #445, opened 2023-05-29
Remove Python 2 support and upgrade requirements.txt

### LINKED ISSUE #439: Remove Python2 support
Now that we're using Python3 for builds here, there's no need to maintain Python2 support. A few things will need to be done:

- [ ] Update requirements.txt so that it has the newest non-broken versions of all the dependencies, to avoid issues with older versions and incompatible dependency chains.
- [ ] Remove any code that has try/catch or if statements or anything for Python2 vs 3 conditional code.
- [ ] Test builds against the contents of tb-website-builds to make sure that there's no unforeseen differences.
- [ ] Update the documentation where necessary.

MelissaAutumn:
Fixes #439

Remove Python2 support

Removed unicode string hint in faq, and a few more python2 -> 3 adjustments

Fixed wsgi.py from incorrectly building the language_map dictionary and a few small formatting errors

Added wsgi pytest file

Upgrade requirements.txt

---

requirements-dev.txt was pretty much spot on. I've adjusted some styling / removed some duplicate lines as well. I diffed again tb-website-builds looks pretty clean! I also ran this through my ide's static analysis tool to catch a few things that may cause issues. 

Verified on Python 3.11.


wsgi.py
```diff
@@ -10,7 +10,7 @@ def get_language_map():
     :return: dict
     """
     LUM = {i.lower(): i for i in settings.PROD_LANGUAGES}
-    langs = dict(LUM.items() + settings.CANONICAL_LOCALES.items())
```
MelissaAutumn, line 13: Was causing a TypeError

product_details.py
```diff
@@ -68,8 +63,6 @@ class ThunderbirdDetails():
 
     minor_releases = load_json('thunderbird_history_stability_releases.json')
 
-    major_releases = load_json('thunderbird_history_major_releases.json')
```
MelissaAutumn, line 71: Duplicate load_json call. This is done before minor_releases on line 67.

helper.py
```diff
@@ -450,13 +448,13 @@ def get_faq_list(ctx):
...
 
 
-@jinja2.contextfunction
+@jinja2.pass_context
 def get_latest_build(ctx, channel):
     """Returns the latest build number for a given channel (e.g. 102.10.0 for release)"""
     return thunderbird_desktop.latest_version(channel)
 
 
-@jinja2.contextfunction
```
MelissaAutumn, line 459: contextfunction was deprecated and can be replaced with pass_context

faq.py
```diff
@@ -11,52 +11,52 @@ def get_entries(ctx):
...
+                gettext('How do I cancel or change my recurring gift?'),
+                gettext('If you used PayPal to set up a recurring gift, you’ll need to <a href="%(paypal)s">log in to your PayPal account</a> to make changes or to cancel. If you made a monthly gift using your credit card, please contact the Donor Care team <a href="%(help)s">via this form</a>. Please include the email address and name you used to make your donation, and they’ll try to help you within two business days.')
                 % {'paypal': 'https://www.paypal.com', 'help': url(ctx, 'thunderbird.donate.contact')}
             ),
             (
-                gettext(u'Who can I email directly with questions about giving?'),
-                gettext(u'If you have a question about giving to Thunderbird, please contact us via <a href="%(help)s">this form</a>. We will do our best to follow-up with you as soon as we can. If you need technical support, please head over to <a href="%(support)s">Thunderbird Support</a> for assistance.')
```
MelissaAutumn, line 59: Py3 is unicode by default

PR #450, opened 2023-06-22
Updated FAQ entries (Fixes #449)

### LINKED ISSUE #449: FAQ Updates
It's that time again, for some FAQ copy updates!

MelissaAutumn:
Fixes #449 

The copy updates requested are finished. 

Updated Copy Preview:
Screen Shot 2023-06-22 at 10 13 53-fullpage

When I first threw these into their own python file I thought it would be easier to wrangle large text, but uhh..I might swap it back to HTML templates in the future.


Sancus:
> When I first threw these into their own python file I thought it would be easier to wrangle large text, but uhh..I might swap it back to HTML templates in the future.

When you first did this I was going say that doing that felt kinda wrong but at the time I couldn't think of a major reason not do it so I just let it go figuring we could always change it back if it ended up problematic later.

Jinja2 does let you set lists as variables and you can also import a variable from a template so I'm pretty sure you can do the same thing with slightly different syntax in a normal template. And then the text will be in the templates folder where it belongs ;)

MelissaAutumn:
I think the main pain point here is editing the text in a non-one giant line that does not wrap way. I RTFM again and went with the trans tag to wrangle that in. That did mean I needed to inline all the entries, but it looks way nicer now. I've also nuked faq.py.

I found most of the faq entries match up (with faq.py) in locale extraction, but there's so many changes here we might need to let this sit on our `master` branch for a few days to allow localizers to catch up.

PR #454, opened 2023-06-27
Ignore `.DS_Store` files when rendering thunderbird-website. (Fixes #453)

### LINKED ISSUE #453: Ignore .ds_store when searching for templates to render
Micah ran into an issue where our build process was trying to pull in .ds_store and erroring out with utf-8 encoding errors. We should just ignore that file

MelissaAutumn:
Resolves #453 

Micah ran into this issue, and it's sort of odd I never did considering I'm also on a Mac. But let's just ignore that specific file for now. We could also just ignore anything that's not an .html file. 

I reformatted the path `startswith` conditions to be more positive, and grouped with the ds_store ignore.

### DIFF
--- a/builder.py
+++ b/builder.py
@@ -248,17 +248,22 @@ class Site(object):
     def render(self):
         """
         Iterate through the templates loaded into Jinja2 and build them, including any needed directories.
-        '_' in front of a template or folder will force those to be skipped by this method.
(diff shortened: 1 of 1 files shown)


Sancus:
I think we should just restrict this to `.html` files only because there's other types of garbage that can be put in there by bad OSes too.

PR #455, opened 2023-06-30
115 release pages first pass

micahilbery:
This is the first pass at the 115 Supernova "What's New" and donation appeal pages.


website/thunderbird/115.0/whatsnew/index.html
```diff
@@ -0,0 +1,289 @@
...
+    </div>
+  </header>
+{% endblock %}
+
+{% block content %}
+  <section class="whats-new-section">
+    <figure class="container slider-box">
+      <div id="before"></div>
+      <div id="after"></div>
+      <input type="range" min="1" max="100" value="50" class="slider" name='slider' id="slider" />
```
Sancus, line 32: This slider is cool(nice work!). When I click on it to move the bar around, it seems like it holds focus, which causes mouse scrollwheel to scroll the bar around instead of the page, until you click somewhere else. That disrupts looking at the rest of the page after interacting with the slider.

Sancus (changes requested):
General comment -- you don't need to change anything, but the {% trans %} block exists to make large blocks of text easier instead of having to wrap them all in `_()`. See [the FAQ page](https://github.com/thundernest/thunderbird-website/blob/master/website/includes/faq.html#L28) for examples.

website/thunderbird/115.0/whatsnew/index.html
micahilbery: What browser / OS? I have tried this on Firefox, Safari, and Chrome on MacOS and I cannot reproduce this. If I click into the slider and use arrow keys it affects the slider but the scrollwheel only scrolls the page for me.

I thought maybe doing a `preventDefault()` onWheel might solve it but that makes it where I can't scroll at all when it's focused.

Sancus: Firefox on Windows has this problem for me, even in Troubleshoot Mode. Chrome does not.

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+  width: 100%;
+  position: relative;
+  isolation: isolate;
+  opacity: 0.75;
+}
+
+.slider-box:hover,
+.slider-box:focus-within {
+  opacity: 1;
+}
```
Alecaddd, line 268: Let's remove all of these.
No need to have opacity on the primary screenshot, it kinda looks bad and it doesn't add anything to the interaction

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+  --focus-outline: 4px solid var(--focus-outline-color);
+  --icon-size: 32px;
+}
+
+body {
+  padding: 0;
+  margin: 0;
+  font-size: 1rem;
+  background-color: var(--bg);
+  color: var(--txt);
```
Alecaddd, line 157: The header image should be on the body so it can scale properly like is in the mock-up
```suggestion
  color: var(--txt);
  background-image: url("/media/img/thunderbird/backgrounds/blob.png");
  background-repeat: no-repeat;
  background-position: center -300px;
  background-size: 80vw;
```

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+}
+
+header {
+  --max-width: 70ch;
+  background-image: url("/media/img/thunderbird/backgrounds/blob.png");
+  background-size: min(150%, 150vh);
+  background-position: bottom;
+  background-repeat: no-repeat;
```
Alecaddd, line 165: We shouldn't have the background image in the header, remove these

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+  padding-inline: 16px;
+  padding-block: 100px 500px;
```
Alecaddd, line 167: Why so huge? Let's try to not force users with normal size monitors to scroll a lot before seeing anything
```suggestion
  padding-block: 50px 400px;
```

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+}
+
+header h1 {
+  font-weight: 300;
+  font-size: 2.5rem;
+  text-shadow: 0 2px 6px var(--color-ink-90);
```
Alecaddd, line 173: The header is a bit messed up.
I'd suggest to use a flex column so we can more easily align the various elements.

```suggestion
  font-weight: 100;
  font-size: 2.5rem;
  text-shadow: 0 2px 6px var(--color-ink-90);
  position: relative;
  display: flex;
  flex-direction: column;
  line-height: 2em;
  margin-block: 30px 0;
```

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+  font-weight: 600;
+  text-transform: uppercase;
+  font-size: 4rem;
+}
+
+.version,
+.alive {
+  color: transparent;
+  font-weight: 900;
+  line-height: 0;
```
Alecaddd, line 186: Never use line-height: 0. It has a loooot of unintended consequences

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+}
+
+.version {
+  font-size: 10rem;
+  background: linear-gradient(30deg, transparent 25%, var(--color-purple-20));
+  -webkit-background-clip: text;
+  background-clip: text;
+  margin-inline-start: -10rem;
+  margin-block-start: -12rem;
+  vertical-align: super;
```
Alecaddd, line 197: If we use flex column, we can do this here
```suggestion
  line-height: 1;
  align-self: end;
  margin-block-start: -9rem;
```

website/thunderbird/115.0/whatsnew/index.html
```diff
@@ -0,0 +1,289 @@
...
+{% block site_header %}
+  <header>
+    <div class="container">
+      <h1>
+        {{_('What’s New in
+        <br>
+        <span class="tb-title">Thunderbird</span>
+        <span class="version">115</span>')}}
+      </h1>
+      <img class="supernova" src="/media/img/thunderbird/whatsnew/115/supernova.png" alt="">
```
Alecaddd, line 22: Write something like "supernova watermark" in the `alt` attribute. Don't leave it empty like we do in Thunderbird.

website/thunderbird/115.0/whatsnew/index.html
```diff
@@ -0,0 +1,289 @@
...
+{% block page_title %}{{ _('What’s New in Thunderbird 115') }}{% endblock %}
+
+{% block site_header %}
+  <header>
+    <div class="container">
+      <h1>
+        {{_('What’s New in
+        <br>
+        <span class="tb-title">Thunderbird</span>
+        <span class="version">115</span>')}}
```
Alecaddd, line 20: Let's add another element here with the work "supernova" and make it visually hidden so the whole header can be read by screen readers

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+  border-radius: 24px;
+  height: 48px;
+  width: 48px;
+  filter: blur(24px);
+  opacity: 0.5;
+  z-index: -1;
+}
+
+.info .header h3 {
+  font-weight: 300;
```
Alecaddd, line 417: Since Open Sans doesn't give much weight flexibility I'd make the titles bold and the description light

```suggestion
  font-weight: 600;
```

media/css/115release.css
```diff
@@ -0,0 +1,1178 @@
...
+
+.slider-tab::after {
+  transform: rotate(-45deg);
+}
+
+.info p {
+  color: var(--info-txt);
+  font-weight: 500;
+  font-size: 2rem;
+  margin-block: 0.5em 1em;
```
Alecaddd, line 381: Let's tweak a bit this to improve readability and visual rhythm.
Also, since all sections have an header with an icon, we can visually indent the paragraph to match the start of the icon.
You can create an exception class in case a section doesn't have a title with an icon.
```suggestion
  font-size: 1.8rem;
  margin-block: 0.5em 1em;
  line-height: 1.35em;
  padding-inline-start: 44px;
```

Alecaddd (changes requested):
Pretty good.
Just some initial tweaks to make it more closer to the mock-ups.
I didn't want to be super nitpicky since we need to land these strings ASAP

website/thunderbird/115.0/whatsnew/index.html
```diff
@@ -0,0 +1,289 @@
...
+      </p>
+    </div>
+  </section>
+
+  <script>
+    const root = document.documentElement;
+    const slider = document.getElementById("slider");
+    slider.addEventListener("input", (e) => {
+      root.style.setProperty("--slider", `${e.target.value}%`);
+    });
```
Alecaddd, line 287: Indeed I can reproduce the issue reported by @Sancus onw Windows.
The problem is that the `input` field steals the focus on click and scrolling on top of the image makes you interact with the slider.

A simple hack we could try is add a `mouseup` event listener to the slider and move the focus back to the body.
Or maybe on `click` we could store the `document.activeElement` and then restore that focus on `mouseup`.
Try to play with these.

website/thunderbird/115.0/appeal/index.html
```diff
@@ -0,0 +1,109 @@
...
+      </div>
+      <div id="hover-hearts">
+        <div id="hover-heart-1">{{ svg('supernova-star') }}</div>
+        <div id="hover-heart-2">{{ svg('supernova-heart') }}</div>
+        <div id="hover-heart-3">{{ svg('supernova-star') }}</div>
+      </div>
+    </a>
+    <div class="letter-container">
+      <p>
+        {{_('Thunderbird’s mission is to give you the best privacy-respecting, customizable communication experience possible.
```
Alecaddd, line 66: This character comes out like this for me.
@Sancus are you seeing this too?

Screenshot 2023-07-06 145740

Sancus: This is a local dev server issue only, can ignore it.

PR #456, opened 2023-07-01
Pull Request Preview Environments

MelissaAutumn:
Woah! I finished this? Wild. 

You can see a preview of this on my local repo:
https://github.com/MelissaAutumn/thunderbird-website/pull/1

This PR brings preview environments to thunderbird-website! Simply attach the `preview environment` label to a PR, and in like 7 minutes you'll have a live instance of that PR. If the PR is updated, the workflow will re-run and generate two new instances and clean up the old ones. If you unattach the label or delete the PR it should also clean up.

Cleaning up is powered by tagging everything in AWS. I create a unique id derived from a particular PR, and use that to lookup specific resources to nuke before generating new instances (or not depending on if we're closing a PR.)

I may have overlooked a specific resource to be cleaned up but we can fix that later. The repository should clean up any images older than 7 days automatically.

I setup our `preview.thunderbird.dev` to wildcard to the load balancer used. And the target groups for that load balancer directs the user to the correct fargate instance. (That was a pain to setup.)

There should be a user setup with the required permissions, but I may have missed one here or there. The task definitions should be created on upload, so don't worry about that. 

I believe I've updated all the hardcoded references, apologies if I didn't.

You'll need to setup the following secrets:
- `AWS_ACCESS_KEY_ID` - Access key of the IAM user
- `AWS_SECRET_ACCESS_KEY` - The secret access key for the IAM user
- `AWS_AVAIL_REGION` - This one is dumb, it's the availability region, so probably us-west-2a or 2b. This lives in the load balancer
- `AWS_LISTENER_ARN` - The HTTPS:443 listener on the load balancer's arn
- `AWS_LOAD_BALANCER_ARN` - The load balancer arn
- `AWS_SUBNETS` - The subnets to use, I was lazy and put this in a secret. This needs to be a list of strings, so grab the two subnets and enter them as `"subnet-arn-1","subnet-arn-2"` (Use double quotes)
- `AWS_SECURITY_GROUPS` - Similar to subnets: `"security-group-1","security-group-2"`, if there's only one then just put one string value. (Use double quotes) 
- `AWS_VPC_ID` - VPC ID
- `PREVIEW_HOST_URL` - `preview.thunderbird.dev`.
- `PREVIEW_IS_SECURE` - `1`

I'll answer any particular questions when I'm able to, but it may be a bit.

### DIFF
--- a/.aws/start-page-task-definition.json
+++ b/.aws/start-page-task-definition.json
@@ -1,0 +1,58 @@
+{
+    "containerDefinitions": [
+        {
+            "name": "thunderbird-start-page",
+            "image": "latest",
+            "cpu": 0,
+            "portMappings": [
+                {
+                    "name": "thunderbird-start-page-80-tcp",
+                    "containerPort": 80,
+                    "hostPort": 80,
+                    "protocol": "tcp",
+                    "appProtocol": "http"
+                }
+            ],
+            "essential": true,
+            "environment": [],
+            "environmentFiles": [],
+            "mountPoints": [],
+            "volumesFrom": []
+        }
+    ],
+    "family": "thunderbird-start-page",
+    "executionRoleArn": "arn:aws:iam::883789023587:role/ecsTaskExecutionRole",
+    "networkMode": "awsvpc",
+    "revision": 2,
+    "volumes": [],
+    "status": "ACTIVE",
+    "requiresAttributes": [
+        {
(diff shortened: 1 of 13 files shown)


bogomil:
That's a pure gold. Can we have this merged please! Thanks @MelissaAutumn

Sancus:
Superseded by #1179 but bird names included :)

PR #465, opened 2023-07-11
download button fix

micahilbery:


Sancus:
This looks good to me. Can we also fix the download button issue Lisa noticed on Chrome as well?

image

micahilbery:
I'm experimenting with stuff now. If I fix it in chrome it's broken in firefox.

micahilbery:
latest commit also fixes #467

PR #487, opened 2023-09-15
Fix zip code

micahilbery:


Sancus:
I was wondering how that would have happened, because I copy pasted what they gave me, but I guess they gave me the wrong zip code.

micahilbery:
It was a mistake at the top of the chain 😅 Everyone was copy/pasting passing it down but it was mixed up the first time it was sent.

PR #497, opened 2023-10-05
Copy over a merchantid file required for Apple Pay during website build.

### LINKED ISSUE #496: Add FRU merchantid file for Apple Pay
Seems like we just need to host their file. Shouldn't be too hard to copy it over during build. 
https://fundraiseup.com/docs/apple-pay/

According to the apple docs (for a similarly named file) they check it once a week, so we can't just verify it and remove it. Boo.

MelissaAutumn:
Fixes #496 

Sort of annoying that we have to host this file, so I just threw it in a misc folder. It gets copied to `thunderbird.net/.well-known/<file name goes here>` on build.

I also added a setting since it's an optional part of this site build but it does default to True.


Sancus:
This is a one-time thing, afaik, so we can change the setting to False by default after the verification is done.

I'm not even sure if it's worth keeping this in the codebase, in the past I've just copied these types of files to the web servers manually and then deleted them after pressing the verify button.

I suppose we may have to verify other subdomains in the future depending on what we do with the website structure.

Sancus:
oh ffs
>A registered domain’s verification expires when its SSL certificate expires.

MelissaAutumn:
lol! Yea I think we'll need to keep this here sadly.

Sancus:
I think this is good, I would add a comment in the settings file that turning that off will break Apple Pay donations, just so nobody ever turns it off unthinkingly.

PR #500, opened 2023-10-06
Display the old download links below the download button on the frontpage

### LINKED ISSUE #483: Can't download for non-default version, system or lanugage anymore
Seems there's no longer any link to https://www.thunderbird.net/en-US/thunderbird/all/ from the site. 
That was very needed, in case you need to download another version to try something out with a version not automatically guessed.

There should also be a link to https://www.thunderbird.net/en-US/thunderbird/releases/ somewhere.

MelissaAutumn:
Fixes #483

A temp solution because these links appear to be quite popular 😄. We can think about a prettier place to stick 'em when we move forward with the rest of the redesign.

I shifted them over to align with the download button. (They both use translate: transformX(-20px);) 

## Before
Screen Shot 2023-10-06 at 11 01 18-fullpage

## After
Screen Shot 2023-10-06 at 11 01 25-fullpage

### DIFF
--- a/assets/css/115-homepage.css
+++ b/assets/css/115-homepage.css
@@ -685,10 +685,6 @@ button:hover::after {
   margin-block: 40px;
 }
 
-p.download-other.download-other-desktop.os_android.os_ios.os_linux.os_linux64.os_osx.os_win.os_win64.os_winsha1 {
-  display: none !important;
-}
-
 .site-nav {
   box-sizing: border-box;
   position: fixed;
@@ -1481,4 +1477,14 @@ footer {
   .btn-newsletter-icon {
     display: inline-block;
   }
+}
+
+.download-other {
+  text-align: center;
+  transform: translateX(20px);
+  margin-top: -1rem;
+}
+
+.small-link {
+  font-size: 80%;
 }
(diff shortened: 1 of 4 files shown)


Sancus:
I would prefer we turn the platform buttons above the button into a link for systems & languages, and I don't think we need the `Privacy` link there.

The `What's New` link doesn't really make sense either since it's using the same string but leads to a completely different place than the "What's New" in the top menu.

MelissaAutumn:
I can see dropping What's New (as that is linked on the System & Languages page), and Privacy, but I think we should still have a "System & Languages" link available. I don't think the little platform icons are suitable for being download links though, too small and not descriptive on their own. 

I can see that working if we embed them into the download button, with a little dropdown to select alternative platforms. (Along with a "More" or a "Other languages" option that links to "System & Languages".)

For now I can just do throw in System & Languages as a temp solution, and we can discuss design at another time/ticket?

MelissaAutumn:
Updated to:
image

Also ugly mock up of what I was suggesting:
image

Sancus:
> I don't think the little platform icons are suitable for being download links though, too small and not descriptive on their own.

To me the fact that it's a combination of the platform icons implies they could be a link to downloads for all platforms, at least that's what I would think if I saw them on another site. Optimally they would have a mouseover effect, though, for sure.

> For now I can just do throw in System & Languages as a temp solution, and we can discuss design at another time/ticket?

Sure that's fine.

PR #520, opened 2023-11-13
Small styling pass for the EOY appeal 2023

MelissaAutumn:
Just some small tweaks I wanted to go done before it sits for a bit longer:
 - Trim the devices image by about 50 pixels to centre it
 - Make the donate button work function
 - Add initial button states for hover and active.

The button states seem like they have different backgrounds as well, however the values in figma are the same. So I just wanted to confirm if the background gradient also needs to change per state, or if they're all the same. 

Screen Shot 2023-11-13 at 12 40 58-fullpage
Screen Shot 2023-11-13 at 12 41 00-fullpage

https://github.com/thunderbird/thunderbird-website/assets/97147377/d236020c-d30b-4335-8229-d460a200af16


micahilbery:
In the Figma I added a 5% opacity white layer on top of the button background to brighten it slightly when hovered and 10% black layer on active to darken it slightly when pressed. This is why the backgrounds look different but the gradient is the same. because the glow and button are separate elements you can probably achieve this with a `filter: brightness()` and not have to mess with the background gradient itself.

MelissaAutumn:
> In the Figma I added a 5% opacity white layer on top of the button background to brighten it slightly when hovered and 10% black layer on active to darken it slightly when pressed. This is why the backgrounds look different but the gradient is the same. because the glow and button are separate elements you can probably achieve this with a `filter: brightness()` and not have to mess with the background gradient itself.

Ah, I've added those tweaks and it looks much more lively. Thanks!

MelissaAutumn:
Here's the updated button states:
https://github.com/thunderbird/thunderbird-website/assets/97147377/284f1cb7-6e2e-4118-813c-b262120a8be8

I swapped the brightness filter out for just directly affecting the rgb values of the colour (which is what the brightness filter does iirc.) I did this so the text's brightness wouldn't be affected by the filter. However we do a bit of smoothness as gradiants can't be animated.

micahilbery (approved):
Thanks for this! LGTM 🚀

PR #531, opened 2023-11-22
Image optimizations for EOY Appeal 2023

MelissaAutumn:
This will hopefully be the last piece of this appeal!

I've adjusted `high_res_img` to allow for a variable scale (defaults to `1.5x`) and a tuple of alternate formats (ordered by tuple position, so if you have `('avif', 'webp')`, the browser will pick 'avif' first.) 

I've also went ahead and converted the images where it made sense (at a small file size of 10kb~ webp wins out over avif) and applied image sets to the background images so they can fall back gracefully as well.

You can't really tell the difference in photos but here they are anyway. Overall a image transfer reduction (no network compression) of 32% (1.25mb -> 403kb~)

Before:
Screen Shot 2023-11-22 at 10 46 53-fullpage

After:
Screen Shot 2023-11-22 at 10 46 38-fullpage


helper.py
```diff
@@ -157,10 +157,29 @@ def high_res_img(ctx, url, optional_attributes=None):
...
     # Use native srcset attribute for high res images
     markup = ('<img class="{class_name}" src="{url}" '
-              'srcset="{url_high_res} 1.5x"'
-              '{attrs}>').format(url=url, url_high_res=url_high_res,
+              'srcset="{url_high_res} {scale}"'
+              '{attrs}>').format(url=url, url_high_res=url_high_res, scale=scale,
                                  attrs=attrs, class_name=class_name)
 
+    # If we've specified some alternate formats we need to use the <picture> tag instead
+    if len(alt_formats) > 0:
```
Sancus, line 165: alt_formats is empty by default so you don't need `len`, `if alt_formats:` should evaluate to False on its own.

MelissaAutumn: Ah good point!

PR #537, opened 2023-12-05
Add 2023 Holiday EOY Appeal.

Sancus:


Sancus:
utm tags should be good so this is ready for review

MelissaAutumn (approved):
I feel like the footer heart should be more vertically centred but I guess that's just how it is!

PR #539, opened 2023-12-10
Fix references to the old GitHub org

babolivier:
I was browsing the website just now and realised the bug report link in the 404 page was still pointing to the thundernest org, so figured I'd fix this and other occurrences. Not a huge deal since GitHub redirects anyway, but might as well.


MelissaAutumn (approved):
Whoops, thanks for catching this!

PR #543, opened 2023-12-19
Add holidayeoy to the `ALWAYS_LOCALIZE` setting, along with a test.

MelissaAutumn:
This should fix up the issue. When I do the site re-write I'll have to make some time to document all of the settings and quirks of this site.

The WSGI server normally doesn't run during dev as we have a separate dev server. I should get that docker PR fixed up and out the door so dev always happens on the WSGI server. 

I've also added a test to help cover some existing paths and ensure this fix actually works before I push it up to stage and then prod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,13 @@ page and thunderbird.net itself.
 
 # Localization
 
+## For Contributors
+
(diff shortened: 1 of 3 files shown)


jfx2006 (approved):
LGTM

PR #546, opened 2024-01-07
adds link to developer.thunderbird.net in holidayeoy

drehuwann:
To allow the non-wealthy users to contribute through their workforce, we should provide a link to our developer network in the call for contributions.


MelissaAutumn:
Thank you for your contribution. We won't be merging this in, but we'll take this feedback into consideration for future appeals.

PR #561, opened 2024-02-22
Reorganize website templates, builds, and external libraries.

### LINKED ISSUE #375: Organize Python Scripts
Right now everything is just dumped into the root directory with a sprinkle of files throughout.

We should consolidate it into `src/` or `tools/` (I'm leaning more on src.) With additional child folders acting as a descriptive and relevant namespace for each file. 

Some suggestions:
Build tools could be in `/src/build`
l10n tools could be in `/src/l10n`
Localizable content (FAQ.py) could be in `/src/text`
General files like `settings.py` and `helpers.py` could just be in `/src`

`build_site.py` and `wsgi.py` can be in the root acting as an entry points. (Alternatively `/bin`)

I've internally slotted this for after the Python3 upgrade.

MelissaAutumn:
Fixes #375

* All website templates are now under the path of `sites/`.
* All external libraries (e.g. product-details, notes, locale) are under `libs/`. (We can change this to git/ or thunderlibs/, etc...)
* Builds are now placed in `dist/`.
* Updated l10n tools, builder.py, translate.py, and settings.py with the appropriate path changes.
* Updated readme.md with path changes.
* Rename start-page to start.thunderbird.net
* Rename website to www.thunderbird.net

I haven't moved any of the actual python script files, that can be a follow-up PR since that will be a little larger in scope due to python import adjustments. But I wanted to get this out before I start work on the re-design.


l10n_tools/extract.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 cd ..
-pybabel extract -F babel.cfg -o locale/templates/LC_MESSAGES/messages.pot .
-cd website 
-cd ../l10n_tools
+pybabel extract -F babel.cfg -o libs/locale/templates/LC_MESSAGES/messages.pot .
+cd sites/website
+cd ../../l10n_tools
```
MelissaAutumn, line 6: I assume I can remove the first cd, but I'd rather not accidentally break something that was working before.

PR #565, opened 2024-04-01
Quarterly calendar update from our calendar provider.

MelissaAutumn:
Sometime in the future I need to add a "has anything changed" check that should just remove the noisy timestamp updates. There's a few additions here though.


Sancus:
The diff is useless because of the timestamp noise but I'm just going to assume this is good, lol.

PR #576, opened 2024-05-07
Filter out versions 125.0, 126.0, 127.0 as they have no release notes - Fixes #575

### LINKED ISSUE #575: Thunderbird 125.0
In the Thunderbird release notes, there is version 125.0. The next page says: the page you are looking for doesn’t exist.

MelissaAutumn:
We could also make this a settings.py value. This should only affect release notes afaik.

### DIFF
--- a/product_details.py
+++ b/product_details.py
@@ -9,6 +9,18 @@ import re
 import settings
 
 
+def filter_major_versions(versions):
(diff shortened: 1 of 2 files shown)


Sancus:
Yeah, I would put versions_to_filter in settings.py. Might as well.

MelissaAutumn:
Annnnd done

PR #596, opened 2024-06-01
Consistently write "macOS"

leoheitmannruiz:
Hey :)

Are these two occurrences supposed to be spelled "MacOS"?

grep -nr "MacOS"


MelissaAutumn:
Ah sorry, I missed this as my notifications settings are weird. Thanks for your contribution!

PR #598, opened 2024-06-06
Show bug numbers in preview mode with feeds template

jfx2006:
The release announce emails use this template, and need to show bug numbers regardless of release type. Code copied over from release-notes.html template.


MelissaAutumn (approved):
Just apply those two fixes and it lgtm!

PR #607, opened 2024-06-24
Fix LinkedIn url (fixes #606)

### LINKED ISSUE #606: LinkedIn link redirects to home page
Hi!

On the participate page (Promote section), the LinkedIn link redirects to the home page.

https://www.thunderbird.net/en-US/participate/

tb-screenshot

The 'participate.promote.linkedin' key is missing on URL_MAPPINGS in settings.py, hence the current rendering.

If you could assign this to me, that would be great!

joetor5:


MelissaAutumn:
Thanks for the patch!

Issue #559, opened 2024-01-29
Initial monthly release download


Sancus:
Eventually, monthly releases will be the "release" build and "esr" will be what is the current release.

For now, this should be an additional build along with nightly and beta.

Product details will need to include these builds for them to be on the website.

Sancus:
Assigned Rob for the product details part.

MelissaAutumn:
Just checking. Was there any movement on this that needs thunderbird-website specific work yet?

jfx2006:
We should think about this a bit but quickly...

115 releases currently update `thunderbird_versions.json` `LATEST_THUNDERBIRD_VERSION` and `THUNDERBIRD_ESR` fields.

128esr releases will update`THUNDERBIRD_ESR_NEXT` **only** until 115 is done at which point it will update `THUNDERBIRD_ESR`.

I'd like to have ShipIt stop updating the `LATEST_THUNDERBIRD_VERSION` for 115 "soon" so that this value can transition to being used for Monthly releases.

At a minimum though, to get 128esr info, product_details.py needs to look at `THUNDERBIRD_ESR_NEXT`.

coreycb:
The PR for updating LATEST_THUNDERBIRD_VERSION to the current monthly is at: https://github.com/mozilla-releng/shipit/pull/1558

PR #613, opened 2024-07-02
Check both Thunderbird_esr_next and Thundebird_esr for the latest version

MelissaAutumn:
Part of #559 

This frees up `LATEST_THUNDERBIRD_VERSION` for use with monthly releases.

I ran into a fun issue where thunderbird_url had a default value of 'None' as a string. So I fixed that and nothing seems out of order.


jfx2006 (approved):
I think this will work.

PR #622, opened 2024-07-15
Serve Windows 7/8.1 users the latest 115 build. (Fixes #621)

### LINKED ISSUE #621: Serve 115 if the user is on Windows 7 or Windows 8
With the release of Thunderbird ESR 128 the minimum requirements are now Windows10+. We shouldn't serve that to Windows 7 or 8 users.

MelissaAutumn:
Main change is the addition of a Windows 7/8.1 download option that detects if they're on Windows NT 6.1+ (Up to 10.0) and shows them `release_win7_8` fake release channel download (which is the latest 115.) It's a little hacky, but it'll do.

For the download buttons we set a platform class and arch sizes on the html document. And there we can reference it without using javascript. Unfortunately the css required can be a bit wacky at times, but I've tested this with a mock useragent and platform set.

### Changes
* Clean up old 115-homepage.css
* Rename win7up class to win10up
* Add win7-8 class
* Add "Windows 7/8.1" option to download options page
* Add "Windows 7/8.1" 32-bit and 64-bit download buttons
* Fix uppercase macOS references
* Adjust platforMap position
* Add override to hide other channels if Windows 7/8.1.


assets/js/common/download.js
```diff
@@ -122,6 +128,18 @@ if (typeof Mozilla === 'undefined') {
     if (firstInstaller) {
       installerSelect.value = firstInstaller;
     }
+
+    // Hack: We need to hide beta and daily for Windows 7/8.1 builds, and force the release channel.
+    if (os === platformMap['windows-7-8']) {
+      document.querySelector('#download-release-select [value="beta"]').classList.add('hidden');
+      document.querySelector('#download-release-select [value="daily"]').classList.add('hidden');
+      // Force release build
+      channelSelect.value = 'release';
```
radishmouse, line 137: classic :joy:

radishmouse (commented):
Hacky and delicious!
Looks reasonable to me :+1:

PR #637, opened 2024-08-08
Fix small error in hero text

Hedreon:
This very small pull request fixes a grammatical error in the hero text.


MelissaAutumn:
Hello! Thanks for the PR, just checked with the team and the copy is intended to be as it is now.

Cheers,
Mel

PR #651, opened 2024-09-24
Adjustments to allow monthly releases (Or "Releases") to be downloaded.

### LINKED ISSUE #559: Initial monthly release download

MelissaAutumn:
* Add a DEFAULT_RELEASE_VERSION and use that for any version defaults.
* Some pre-built metadata for javascript.
* Add a warning about Thunderbird Releases.

Fixes #559 

I'm not sure why I wasn't using a pre-built javascript metadata script tag, but now I am. Other random script tags we have will slowly migrate to that file. 

Here I've split release into the in-testing monthly releases, and esr which will server...esr builds. I've also added a default setting so we can flip it to release in the future with hopefully minimal issues. 

### Screenshots

image
image
image

### DIFF
--- a/build-site.py
+++ b/build-site.py
@@ -57,7 +57,8 @@ elif args.downloadlegal:
 else:
     print('Rendering www.thunderbird.net ' + langmsg)
     # Prepare data and build main website.
-    version = helper.thunderbird_desktop.latest_version('release')
+    default_channel = settings.DEFAULT_RELEASE_VERSION
+    version = helper.thunderbird_desktop.latest_version(default_channel)
     beta_version = helper.thunderbird_desktop.latest_version('beta')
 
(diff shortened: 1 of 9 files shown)


build-site.py
```diff
@@ -79,7 +80,8 @@
                'CALDATA_URL': settings.CALDATA_URL,
                'latest_thunderbird_version': version,
                'latest_thunderbird_beta_version': beta_version,
-               'blog_data': feedparser.parse(settings.BLOG_FEED_URL)
+               'blog_data': [],
```
MelissaAutumn, line 83: We don't use blog_data anymore, so we don't need to poke the blog everytime we build.

PR #671, opened 2024-10-16
Copy 128 update appeal to 115 and change campaign name.

Sancus:


MelissaAutumn:
If it renders fine, it looks good to me!

PR #694, opened 2024-11-15
Add readthedocs

malini:
We're moving toward standardizing the docs we generate in web services, so let's start here as we have some docs to write for how to update TB websites.

You can see the generated docs here: https://thunderbird-website.readthedocs.io/en/latest/

Right now, it's just using the .md files in the docs/ folder, we can hook up python autodoc (or other tool) to generate API docs after

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,8 +3,8 @@
 # -- Project information
 
 project = 'Thunderbird Websites'
-copyright = '2024, Mozilla'
-author = 'Mozilla'
+copyright = '2024, MZLA Technologies Corporation'
+author = 'MZLA Technologies Corporation'
 
 release = '0.1'
 version = '0.1.0'
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -2,9 +2,9 @@
 
 # -- Project information
 
-project = 'Lumache'
-copyright = '2021, Graziella'
-author = 'Graziella'
+project = 'Thunderbird Websites'
+copyright = '2024, Mozilla'
+author = 'Mozilla'
 
 release = '0.1'
 version = '0.1.0'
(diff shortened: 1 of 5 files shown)


malini:
NOTE: once this gets merged, I'll have to update the build branch used in https://thunderbird-website.readthedocs.io/en/latest/ to point to main instead of this branch

docs/conf.py
```diff
@@ -0,0 +1,40 @@
+# Configuration file for the Sphinx documentation builder.
+
+# -- Project information
+
+project = 'Lumache'
```
MelissaAutumn, line 5: these should be adjusted.

malini: ty!

malini: I've edited and regenerated the docs, though you may need to clear cache to see the difference

docs/conf.py
```diff
@@ -0,0 +1,40 @@
+# Configuration file for the Sphinx documentation builder.
+
+# -- Project information
+
+project = 'Thunderbird Websites'
+copyright = '2024, Mozilla'
```
MelissaAutumn, line 6: Not MZLA Technologies Corporation?

docs/conf.py
```diff
@@ -0,0 +1,40 @@
...
+author = 'Mozilla'
```
MelissaAutumn, line 7: Maybe Thunderbird or Thunderbird Services would be better here!

MelissaAutumn (commented):
Can you add a link to the readthedocs in the readme.md?

docs/conf.py
malini: I can open a related ticket in the atn repo later

PR #696, opened 2024-11-17
Fix 115 nov24 appeal url and remove duplicate template code. (Fixes #695)

### LINKED ISSUE #695: Copy Nov24 appeal to 115 folder
Also change the UTM tag a bit so we can see if there's any difference in donations between 128 and 115, while we're at it.

Sancus:


MelissaAutumn (approved):
LGTM!

PR #710, opened 2024-11-23
Adjust background for certain breakpoints.

radishmouse:
For some breakpoints, there was a gap between the background image and the background gradient.

This PR makes the gaps go away.


Sancus:
Please put context in your commit messages, it wasn't clear which background you were referring to. I've force pushed master to fix it so you'll probably need to pull before continuing :)

radishmouse:
> Please put context in your commit messages, it wasn't clear which background you were referring to. I've force pushed master to fix it so you'll probably need to pull before continuing :)

Yes! Sorry about that. Thanks for fixing :+1:

PR #714, opened 2024-11-25
Add fr and de hero images (Fixes #713)

### LINKED ISSUE #713: Update the hero image with additional localizations
Incoming fr and de from freehive.

MelissaAutumn:
Fixes #713 

Pinging @jobisoft and @babolivier for feedback as I do not speak German, and my highschool french only gets me so far. (No rush on the feedback!)


babolivier:
French looks good to me, translations seem to match what's in Pontoon. A couple of nitpicks on the label names:

* the translation for "Todo" is a bit awkward (it would translate to "do" rather than "to do"), "À faire" would be more fitting
* the translation for "Upgrade" is correct, but I think "Mise à jour" would be a tiny bit more natural

jobisoft:
The german translation "Lass uns fliegen! Es ist Zeit für Thunderbird 128." for "Let's fly! It's time for Thunderbird 128." is a literal translation, word by word. But the german translation does not seem to have the intended meaning. What are you trying to express?

* Let's go! It is time for Thunderbird 128. -> "Lass uns loslegen! Es ist Zeit für Thunderbird 128."

@kaie : What do you think?

PR #720, opened 2024-11-26
Fix gradient, fix donation url, and adjust for l10n title text. 

### LINKED ISSUE #719: Fix gradient on dec24 appeal

MelissaAutumn:
Fixes #719 

Re-worked the gradient to be a small block covering the header so that it scales with the screen size. Adjusted the body to have a max-width of 1280px (in rem), and fixed some text clipping with the appeal button on certain locales. 

Also fixed the donation button url!

### DIFF
--- a/assets/less/appeals/dec24.less
+++ b/assets/less/appeals/dec24.less
@@ -45,81 +45,36 @@
 
 
   body {
-    margin: 0;
+    max-width: 80rem;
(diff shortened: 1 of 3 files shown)


radishmouse (approved):
Looks great! Just a note: for viewports width between 1024px and 1280px, does the header-gradient sit too far below the background image?

PR #770, opened 2025-02-28
Appeals Redirect

MelissaAutumn:
I'll cherry pick this on prod after we're done here. 

This adds a redirect from updates.thunderbird.net/<lang>/thunderbird/appeal -> updates.thunderbird.net/<lang>/thunderbird/128/mar25/donate/

The redirect url is adjustable in settings for ez access. I've also copied over the dec24 appeal as mar25, however the FRU form.

### DIFF
--- a/assets/less/appeals/dec24.less
+++ b/assets/less/appeals/dec24.less
@@ -40,7 +40,7 @@
 }
 
 // Scope to just the appeal page
-.page-appeal-dec24, .page-appeal-mar25 {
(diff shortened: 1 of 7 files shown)


malini (approved):

MelissaAutumn:
Talked to lisa, it's actually apr25 appeal. I'm not sure if the time was moved ahead, or if it's needed for march for beta testing reasons, but I've adjusted it and added the FRU id.

PR #793, opened 2025-04-14
Add donation notice to inform user if FRU is blocked (within 10 secs) (Fixes #792)

### LINKED ISSUE #792: Donation notice
Some addons block fundraise up. We should show a notice informing them that is happening and what they can do to prevent it.

MelissaAutumn:
This adds a check if they've clicked on a donate button, or are on the donate page that if fundraise up hasn't loaded within 7.5 seconds (tweakable) then to show a notice (shown in the screenshots below.) 

This only shows up on user action (with the exception being the donate page which starts immediately on page load.) and does not show if fundraise up's javascript checkout event triggers, or if the page finds the embedded donate page widget.

## Screenshots
image
image
image


inattee:
Could we polish up the message a bit:

Your browser or an extension might be blocking our donation platform, Fundraise Up.
Please try again using private browsing (incognito mode) or switch to another browser.

(I don't think it's necessary to include "The Thunderbird Team")

MelissaAutumn:
Thanks I've applied the copy tweaks:
image

PR #813, opened 2025-05-14
Fixup product_details 128esr not being major

### LINKED ISSUE #807: 128esr is not showing up on the releases page
128.0esr isn't showing up on the releases page.

https://www.thunderbird.net/en-US/thunderbird/releases/

Release notes does exist for it so it's probably a thunderbird-website's error.

https://github.com/thunderbird/thunderbird-notes/blob/master/esr/128.0esr.yml

janbrasna:
Fixes #807

At https://github.com/mozilla-releng/product-details/commit/eb4828d4640095484227d88863d316f6e533d320#r156969608 the 128esr became major, but there's no 128* entry in major listing.

Not having major = not being shown. This adds it to esr fixup, to treat it as before.

### DIFF
--- a/product_details.py
+++ b/product_details.py
@@ -212,8 +212,8 @@ class ThunderbirdDetails():
 
         Quirks:
         - 38.0.1 starts with a point release and is added twice, once as a major and second as a stability release.
-        - 115 releases are patched up to be ESR builds.
-        - non-esr builds post 115 are ignored.
+        - 115 releases are patched up to be ESR builds, non-esr builds post 115 are ignored.
+        - 128.0esr has changed from esr to major, but list of majors has no 128 entry.
         """
         releases = {}
 
@@ -230,9 +230,11 @@ class ThunderbirdDetails():
 
         def needs_esr_fixup(version_ints: list[int]):
             """115.10.2 up until 128.0esr are mislabelled and should be esr builds"""
+            if version_ints[0] == 128:
+                return True
+
             if version_ints[0] != 115:
                 return False
-
             # If >=115.11
             if version_ints[1] >= 11:
                 return True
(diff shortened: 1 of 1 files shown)


product_details.py
```diff
@@ -230,9 +230,11 @@ def needs_major_fixup(version_ints: list[int]):
 
         def needs_esr_fixup(version_ints: list[int]):
             """115.10.2 up until 128.0esr are mislabelled and should be esr builds"""
```
MelissaAutumn, line 232:
```suggestion
            """115.10.2 to 128.0esr are mislabelled and should be esr builds"""
```

MelissaAutumn (approved):
One comment change, but this looks good. 

There's another line:
```
is_esr = (category == 'esr' and version_int[0] >= 128) or needs_esr_fixup(version_int)
```

That we could probably drop the >= and make it a > but it works as-is. 

Thanks for the PR!

janbrasna:
I did check again for good measure the fixup impact wouldn't affect something like:

https://github.com/thunderbird/thunderbird-website/blob/8f3c209f8f9ae290446f03446e055d9bcc80d877/product_details.py#L267-L268

turning that into `123.4.5esresr` or something, but the listing looks as expected (I guess?):

> Screenshot 2025-05-15 at 12 26 05

product_details.py
janbrasna: I thought about that `is_esr` bool mentioned above and while true that now it's matching both the conditions at the same time, I'd err on the side of keeping it as-is in case the upstream data issue gets resolved. The rule and its comment there kinda make sense, so I wasn't comfortable changing it too much, in case this fixup can be removed.

A _product-details_ fix may come in a form of adding a major to the majors json at least for now, so the fixup may need to stay even after that… but, wanted to make it reversible with as little foot guns as possible;)

(Appreciate the wording update here, I tried three different descriptions making more obvious what version ranges it concerns, but none sounded good — thanks for helping out with language! You know best what needs to go there to make sense to the next person looking at it;)…)

PR #815, opened 2025-05-18
ImageOptimize PNGs

janbrasna:
There were a few PNGs in recent updates that looked like could use a little optimization, and upon looking closely there seemed to be a few more assets with their file size rather large…

I ended up running most of PNGs eventually through @ImageOptim using OptiPNG / OxiPNG, Zopfli & PNGOUT that all should be lossless, meaning no change in quality whatsoever, only storage & bandwidth saved — approx 15MB out of the 50MB processed.

There were some large images with 50% savings, there were some normal size images also getting 50% smaller so it was worth it; generally the savings are around 30% in average.

(There were some extremely large assets as blob/nebula graphics, or team photos etc. that would need processing manually to save some space, but they seem no longer linked from the site so I skipped spending time on them. I may however tackle the `favicon.ico` at some point separately and encode each variant first, as that one could also be a third its size if optimized a bit.)

There are some interesting savings in landing screenshots, base backgrounds, body header, logo or background, couple hundred kB from contributors graphics, the testimonial hex has great savings, couple hundred kB in total for features graphics and icons, logos, favicons  etc. that's loaded on many common places of the site.

_(I'm not familiar with how the caching of static assets is handled, how to not necessarily invalidate whatever's being cached and delivered via CDNs currently, to kinda let the downstream refresh naturally… as it's probably just not worth invalidating for no meaningful change here.)_


MelissaAutumn:
Thanks for the PR! Looking through them all I don't see any visible difference. We mainly serve webp/avif so most folks won't even get the png versions, but it's good to reduce their size. 

We use cloudflare for caching the website, it'll be cleared automatically after a day or so.

PR #820, opened 2025-05-19
Fix product hero spacing (covering heading on mobile)

### LINKED ISSUE #819: Product hero heading covered by nav on mobile
Home and other pages are OK, only the product (desktop/mobile) header treatment makes the title heading tuck under the nav:

fxios

_(Maybe it was so-so okay~ish before the channel banner, but with it added now the menu just covers the masthead.)_

janbrasna:
Fixes #819

Using the correct custom property, the rest of the breakpoint adjustments can be removed and just left to the same math to work it out.

### DIFF
--- a/assets/less/pages/home.less
+++ b/assets/less/pages/home.less
@@ -18,7 +18,7 @@
     background-position: bottom center, center;
     background-repeat: no-repeat;
     color: var(--color-gray-20);
-    padding-top: var(--nav-size);
+    padding-top: 0;
     isolation: isolate;
   }
 
--- a/assets/less/pages/products.less
+++ b/assets/less/pages/products.less
@@ -32,13 +32,13 @@
     background-repeat: no-repeat;
     color: var(--color-gray-20);
     isolation: isolate;
-    padding-top: var(--nav-size);
+    padding-top: 0;
 
     // More we take from the bottom, the more we need to pull back up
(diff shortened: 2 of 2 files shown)


assets/less/pages/home.less
```diff
@@ -18,7 +18,7 @@
     background-position: bottom center, center;
     background-repeat: no-repeat;
     color: var(--color-gray-20);
-    padding-top: var(--nav-size);
+    padding-top: 0;
```
janbrasna, line 21: NB: This also explicitly removes the non-existent custom property from home to make it more obvious — as using such value kinda unintuitively doesn't just ignore it, using initial/inherited value, but interpolates its falsy to "0" — so instead of the global 10rem for `#masthead`/`‹header›`, this intention to change it to 5rem for the var actually makes it 0px…;)

It does however seem that the current homepage layout works with the 0px padding, as fixing this to the 5rem var would make the hero "too heroic" (:) and occupy too much space. If that's intended I'm happy to follow up separately, now it's just explicitly assigning the value it's effectively setting currently anyways, to make it clear, and to remove the undefined property.

MelissaAutumn: Looks fine to me, thank you!

MelissaAutumn (approved):
Thanks for the PR! Looks good to me 😄

PR #824, opened 2025-05-20
Update Mozilla branding

### LINKED ISSUE #765: Update branding
Mozilla logo and branding was updated. We'll need to update the logo on our websites. It should take up about the same amount of space as the current logo. (Unlike how it is on mozilla.org.) It should also affect all pages including donation appeals (which should be automatic if you catch all the references.)

Might have to adjust the svg to allow fill by [currentColor](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color). 

Image

janbrasna:
Resolves #765

Logo updated to current lockup.

WIP: aspect ratio & whitespace impact to tweak(?)


sites/www.thunderbird.net/includes/appeal-footer.html
```diff
@@ -1,5 +1,5 @@
 <div class="mzla">
-  <a href="https://mozilla.org" class="mozilla-logo">
+  <a href="https://www.mozilla.org" class="mozilla-logo">
```
MelissaAutumn, line 2: I guess this should pull from a url('mozorg.home')

sites/www.thunderbird.net/includes/base/footer.html
```diff
@@ -44,7 +44,7 @@ <h4>{{ _('Contribute') }}</h4>
       </div>
     </div>
     <div class="mzla">
-      <a href="https://mozilla.org" class="mozilla-logo">
+      <a href="https://www.mozilla.org" class="mozilla-logo">
```
MelissaAutumn, line 47: Same for url('mozorg.home')

MelissaAutumn (commented):
Thank you for the PR!

If you have some spare cycles could you left align the mozilla logo at the bottom as mentioned in the issue? (Basically like how https://blog.mozilla.org/en/ has it)

I'll do check out the PR for old appeals later today or early tomorrow (UTC-7).

Thanks again!

janbrasna:
Now the older footer occurrences are also left-aligned with a new 🏴 logo.

MelissaAutumn:
I will say @radishmouse has been handling branding updates for stage.tb.pro. Might be good to consolidate any differences from this PR!

janbrasna:
I'm following the tbpro and common includes, and making this PR deliberately avoiding any overlap — as it's mostly for preexisting content.

I also have a separate change (a few lines) to streamline both wtn+utn with tbpro+common once this merges, it's literally just this d56ff6f +dedupe assets, so the current efforts are well isolated for now, and super easy to blend together once convenient.

janbrasna:
(Also, re: differences — I too coincidentally had the logo sized to 150px wide originally as tbpro+common now, but later changed to 10rem for consistency with other things in the layout. Maybe that would be a good addition to the tbpro styles later on, if there's more things around it remified? /haven't checked there myself too closely yet/…)

sites/www.thunderbird.net/includes/base/footer.html
janbrasna: FWIW these are now moved to the common footer — so I'll rebase that conflict out — but I still think this PR could be useful to clean up the old branding from updates/pledges pages and bump these 115–140 era templates to the current logo as well, to get rid of the old asset at least;D so I'm not closing it for now, even though for the main site it has been superseded by the new shared footer at this point.

PR #828, opened 2025-05-23
Add underlines to major versions in releases

### LINKED ISSUE #826: New full versions' (.0 releases) release notes are not showing on the website as links
On the main "Releases" website https://www.thunderbird.net/en-US/thunderbird/releases/, all the ".0" releases are not showing. Such as 138.0, 137.0 etc. They can still be accessed by going to their exact url (such as https://www.thunderbird.net/en-US/thunderbird/138.0/releasenotes/), but they don't show as clickable links in the main "Releases" website.

Image

janbrasna:
Resolves #826

The fancy heading style makes the default link decoration transparent as a byproduct. Shouldn't be an issue generally, but in the release list it could look like the major versions are missing, so this adds some preexisting subtle affordance (from transparent to matching background enables it to acquire some shadows et al. for light decoration) to imply the headings are actual links to that versions.

### DIFF
--- a/assets/less/pages/releases.less
+++ b/assets/less/pages/releases.less
@@ -26,6 +26,15 @@
     min-width: 15%;
     margin-top: 0;
     text-align: left;
+
+    a {
+      text-decoration-color: var(--bg);
+      text-decoration-thickness: 1px;
+      text-underline-offset: 0.2em;
+      &:hover {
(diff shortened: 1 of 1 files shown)


MelissaAutumn (approved):
Perfect, thank you!

PR #829, opened 2025-05-23
WIP: initial poc of switching to holidays python library

### LINKED ISSUE #753: use free software to generate public calendars
Hi!

I'm not sure this is the right place for this request, but I have found out that the Thunderbird website ships world calendars, thanks to Nextcloud suggesting those calendars in its user interface:

https://www.thunderbird.net/en-US/calendar/holidays/

While trying to figure out where that data was generated, I stumbled upon this project which, in its README file, says:

> Calendar generation can be manually built by appending the option`--buildcalendar`. This queries our current calendar provider (Calendarific) and generates a `.ics` file per each locale specified in settings.py. For testing, you can limit this to just US by using the option `--enus`. 
> This option requires setting the `CALENDARIFIC_API_KEY=` environment variable. If you're using a paid plan you can also set `CALENDARIFIC_IS_FREE_TIER=false` to remove the sleep time between calls.

Unfortunately, [Calendarific](https://calendarific.com/) doesn't seem to be providing or based on free or even open data. In fact, their [terms of service](https://calendarific.com/terms) seem to indicate Thunderbird might actually be in breach of the agreement, as it infringes on point 4.2.e:

> 4.2. Permitted Use. 
> [...]
> The following is a non-exhaustive list of practices that would not be considered "Legitimate Use":
> [...]
> (e) Redistribute or resell the Data or the Services (wholly or in part).

I have not reported this to the Calendarific people, of course.

I have found two Python libraries that provide similar data:

- [Holidays](https://github.com/vacanza/holidays/): [no ICS output](https://github.com/vacanza/holidays/issues/2179), but [pending PR](https://github.com/vacanza/holidays/pull/2261)
- [Calendra](https://github.com/jaraco/calendra): [built-in ICS output](https://workalendar.github.io/workalendar/ical.html)

I encourage you to consider switching to one of those for generating the ICS files! It would reduce legal exposure to the Thunderbird project, at least.

I also think the data source should be credited on the website. If you switch to those free libraries, it could also encourage collaboration in the wider "time and date" community and possibly convergence over such tools and data sources, reducing overall maintenance burden.

Thank you for your consideration.

eyJhb:
This is just an initial POC of how it could look like, if we switched to the Holidays library in Python. It works fairly well, and it seems to be well maintained, with an active userbase.

The library in question https://github.com/vacanza/holidays

This PR needs a lot of cleanup. I just threw it together, to get some idea of how everything works together. But this should be fairly doable? However, the calendar will look vastly different from the old one, and there are a lot more types in this calendar (bank, government, school, etc. etc.).

This would close #753 , if it gets implemented fully.

### DIFF
--- a/calgen/providers/VacanzaHolidaysProvider.py
+++ b/calgen/providers/VacanzaHolidaysProvider.py
@@ -1,0 +1,55 @@
+from calgen.models.Calendar import Calendar, CalendarTypes
+
+from calgen.providers.Provider import Provider
+
+import holidays
+
+
+class VacanzaHolidaysProvider(Provider):
+    def __init__(self, auth_options: dict):
+        super().__init__("VacanzaHolidays", {})
+
+    def query(self, country: str, year: int, additional_options: dict) -> list:
+        if country == "US":
+            print(country, year, additional_options)
+        """Queries Calendarific, will return either the response data, or None if the api returns garbage data."""
+        if country is None:
+            raise RuntimeError("Country parameter is missing")
+        if year is None:
+            raise RuntimeError("Year parameter is missing")
+
+        language = additional_options.get("language")
+
+
+        calendar_type = additional_options.get("calendar_type")
+        language = additional_options.get("language")
+
+        if calendar_type is None:
+            raise RuntimeError("Calendar Type additional option is missing")
+
+
+        try:
+            cal = holidays.country_holidays(country, years=year, language=language)
+        except NotImplementedError:
+            print(f"no calendar available for {country} with language {language}")
+            return []
+
+        return cal.items()
+
+    def build(self, country: str, year: int, additional_options: dict):
+        """Queries Calendarific, builds, and returns a list of Calendarific models from the queried data."""
+        holidays = self.query(country, year, additional_options)
+
+        holidayModels: list[Calendar] = []
+        for d, n in holidays:
+            c = Calendar()
+            c.unique_id = hash(str(d) + n)
+            c.year = year
+            c.iso_date = d
+            c.name = n
+            c.description = n
+            c.calendar_type = CalendarTypes.NATIONAL
+
+            holidayModels.append(c)
+
+        return holidayModels
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -27,3 +27,4 @@ pytest~=7.3.1
 python-dateutil~=2.8.2
 packaging~=23.1
 setuptools==78.1.1
+hollidays>=0.73
--- a/build-site.py
+++ b/build-site.py
@@ -12,6 +12,7 @@ import settings
 from datetime import date
 
 from calgen.providers.CalendarificProvider import CalendarificProvider
+from calgen.providers.VacanzaHolidaysProvider import VacanzaHolidaysProvider
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--enus', help='Only build the en-US language.', action='store_true')
@@ -66,7 +67,8 @@ elif args.buildcalendars:
     except KeyError:
         sys.exit("No `CALENDARIFIC_API_KEY` defined.")
 
-    build_calendar.build_calendars(CalendarificProvider({'api_key': api_key}), calendar_locales)
+    # build_calendar.build_calendars(CalendarificProvider({'api_key': api_key}), calendar_locales)
+    build_calendar.build_calendars(VacanzaHolidaysProvider({}), calendar_locales)
 elif args.downloadlegal:
     print("Downloading legal documents")
     legal = builder.Legal(settings.WEBSITE_PATH)
--- a/build_calendar.py
+++ b/build_calendar.py
@@ -107,8 +107,8 @@ def build_calendars(provider: Provider, locales: dict):
         for country_name, language_code in country_info_list:
 
             # Wait 1 second due to free api restrictions
-            if is_free_tier:
-                time.sleep(1)
+            # if is_free_tier:
+            #     time.sleep(1)
 
             ical = build_ical(provider, locale, language_code, years_to_generate)
 
--- a/calgen/models/Calendar.py
+++ b/calgen/models/Calendar.py
@@ -47,8 +47,8 @@ class Calendar(object):
         data = {
             'uid': self.unique_id,
             'last-modified': datetime.now(),
-            'dtstart': self.iso_date.date(),
-            'dtend': self.iso_date.date() + timedelta(days=1),
+            'dtstart': self.iso_date,
+            'dtend': self.iso_date + timedelta(days=1),
             'summary': self.name,
             'description': self.description,
             'dtstamp': datetime.now(),


calgen/providers/VacanzaHolidaysProvider.py
```diff
@@ -0,0 +1,109 @@
...
+            cal = holidays.country_holidays(country, years=year, language=language)
+        except NotImplementedError:
+            print(f"no calendar available for {country} with language {language}")
+            return []
+
+        # all subdivisions for this country
+        subdivs = cal.subdivisions
+
+        # get the names for all the subdivisions
+        subdiv_to_name = {v: k for k, v in cal.subdivisions_aliases.items()}
```
MelissaAutumn: I think you can probably drop retrieving the aliases. 

It doesn't look like we could reverse this dict as at least in Canada aliases include the French names for Provinces but no information on what locale the name might be from. (https://holidays.readthedocs.io/en/latest/auto_gen_docs/canada/#holidays.countries.canada.Canada.subdivisions)

calgen/providers/VacanzaHolidaysProvider.py
```diff
@@ -0,0 +1,109 @@
...
+        holidays_with_subdivs: dict = {}
+        for subdiv in subdivs:
+            tcal = holidays.country_holidays(
+                country,
+                years=year,
+                language=language,
+                subdiv=subdiv,
+                categories=cal.supported_categories,
+            )
+            for d, n in tcal.items():
```
MelissaAutumn: nit: I think we can write out date and name, it's only a few more letters 😄

calgen/models/VacanzaHolidays.py
```diff
@@ -0,0 +1,59 @@
...
+    ):
+        """Handle regional naming for holidays that fall under the CalendarTypes.LOCAL"""
+        # there is no subdivs -> everywhere
+        if not len(subdivs):
+            return name
+
+        # this should really be localised, but uncertain how to do this best
+        subdivs_except_text = ""
+        if subdivs_except:
+            subdivs_except_text = "All except "
```
MelissaAutumn: I think this would be okay to localize ourselves, so you can probably pull in gettext_object from translate.py and pass in the locale. It should return an initialized gettext instance so you can use the return value like:

```
regions = ', '.join(subdivs)
translator.gettext('All except %(list_of_regions)s' % {'list_of_regions': regions})

Which might produce: All Except BC, ON, QC
```

Unfortunately we use old-style python string formats for l10n. The localizers will see this as just `All except %(list_of_regions)s` so they can place it as needed.

janbrasna (commented):
FYI I did generate the actual .ics too, https://github.com/janbrasna/thunderbird-website/commit/fe68215 — in case we want to show folks the new data and check how viable it is for their regions…

One thing to add to TODO: no need to require the API key if using this new provider.

One nit spotted:

eyJhb:
i've tried to clean up the code a bit (not pushed yet), but most of it ended up in the model instead. I still think it looks cleaner in the end, but I'll remove a bunch of more stuff.

I'll try to get it pushed tomorrow at some point :) It shouldn't change the output, EXCEPT that I'm now skipping any calendars with NO events. Otherwise we would produce empty calendars that would error out the parser AFAIK. But this should maybe be changed :)

eyJhb:
I've pushed the changes now, and there are quite some. I also tried to update the `holidays-calendars.md` file a bit, or at least remove calendarific references. 

I've moved most of the logic into the `VacanzaHoliday` model instead, to make it more simple and clear. It's not really that pretty, any of the code, but it's hopefully not impossible to read.

I've tried to not touch too much of the `Calendar` things, e.g. `from_api`.

eyJhb:
Throwing in a tiny bump here :)

requirements-dev.txt
```diff
@@ -27,3 +27,4 @@ pytest~=7.3.1
 python-dateutil~=2.8.2
 packaging~=23.1
 setuptools==78.1.1
+holidays>=0.73
```
janbrasna, line 30: Let's bump to:

```suggestion
holidays>=0.83
```

that has all the required regions we need.

janbrasna: +QQ: passing tests?

janbrasna:
This PR has currently a conflict with master that needs rebasing, but hopefully it's just some easy readme or config.

While I think removing the old model and provider is called for, as there's not really a reason to go back, there are tests relying on some of its specifics. If this approach is (pre-)approved by the maintainers, the tests would also need updating for the new data.

janbrasna:
QQ: … and I've also seen it in tests — is it expected to have the holidays sets as `OPAQUE` i.e. as "busy" by default? Or that's a feature, for the public/national/bank holidays per se? (And in case there would be other types of holiday events added/configured like observance or religious, those would conversely be configured as `TRANSPARENT` explicitly for the difference?)

docs/holiday-calendars.md
```diff
@@ -24,13 +21,7 @@ Once a quarter the calendars need to be regenerated to ensure accuracy. You can
...
 ```
 
-You'll want to ensure you have the following env vars set:
-
-```shell
-PYTHONUNBUFFERED=1 ;CALENDARIFIC_API_KEY=<your key>;CALENDARIFIC_IS_FREE_TIER=false
-```
-
-If you are an MZLA employee you can find the api key within the infra vault in 1password.
+Please ensure that you are using the latest release of the hollidays library.
```
MelissaAutumn, line 24:
```suggestion
Please ensure that you are using the latest release of the holidays library.
```

calgen/providers/VacanzaHolidaysProvider.py
```diff
@@ -0,0 +1,82 @@
...
+        for subdiv in subdivs:
+            tcal = holidays.country_holidays(
+                country,
+                years=year,
+                language=language,
+                subdiv=subdiv,
+                categories=cal.supported_categories,
+            )
+
+            map_subdiv_to_name = {v: k for k, v in cal.subdivisions_aliases.items()}
```
MelissaAutumn, line 50: This doesn't seem to work for Canadian holidays, as our aliases include the french name for it as well. However not all provinces have an official french name like Manitoba. 

`('British Columbia', 'BC'), ('Colombie-Britannique', 'BC')`

leading to the `BC => Colombie-Britannique` overriding the english name

which ends up with  `SUMMARY:British Columbia Day (Colombie-Britannique)`.

I'm not sure if there's a good fix here, might be an issue we would need to raise with the library.

MelissaAutumn:
^ I forgot I had half a review ready from May :facepalm:

calgen/models/VacanzaHolidays.py
```diff
@@ -0,0 +1,92 @@
...
+            excluded_subdivs: list[str] = []
+            for subdiv in self._all_subdivs:
+                if subdiv in subdivs:
+                    continue
+
+                excluded_subdivs.append(subdiv)
+
+            # insert the excluded text here
+            # it should be done in a better way than this...
+            excluded_text = "All except "
```
janbrasna, line 64: Ah, I was surprised where the "Año Nuevo Judío (Rosh Hashana) (All except Tucumán)" comes from in those instances;)

QQ: If that is wrapped in gettext… does the calendar generation actually run in any l10n context, to apply anything besides just the en-US defaults for such strings?

EDIT: Now I see the comment above, where Mel describes how to add an initialized instance — nice!

PR #849, opened 2025-06-12
Keep track of which site we're building, and transform some relative urls to absolute urls. (Fixes #848)

### LINKED ISSUE #848: Fix relative thunderbird urls when called outside of tb website

MelissaAutumn:
Now that we're using thunderbird. url keys outside of thunderbird website we need a way to transform them to absolute urls. So we'll just keep track of what site we're building and within the url helper use that information to prepend the canonical url if we're outside of tb website.

### DIFF
--- a/builder.py
+++ b/builder.py
@@ -58,14 +58,14 @@ def mkdir(path):
             raise
 
 
-def write_site_htaccess(renderpath: str, lang: str, redirects: dict, redirect_request=False):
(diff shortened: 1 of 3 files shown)


Sancus:
The only thing that seems awkward about this is that it produces links to the live sites when you're running a dev server, so you can't actually follow them. This doesn't matter for the particular case we have now but it could in the future if you're using more of those URLs and you're trying to test.

I am good to merge this as-is and we can resolve that later.

PR #852, opened 2025-06-15
Expose download buttons to l10n

### LINKED ISSUE #851: Download button not exposed to localization
Image

janbrasna:
Fix #851

Main download CTA and desktop menu button translated with gettext, terms already existing.

(Accessibility label removed now when the implicit content is the same.)

### DIFF
--- a/sites/www.thunderbird.net/includes/base/nav.html
+++ b/sites/www.thunderbird.net/includes/base/nav.html
@@ -80,8 +80,8 @@
   </li>
   <li class="nav-entry no-border no-mobile nav-button-fix">
     <div class="header-download">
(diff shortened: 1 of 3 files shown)


janbrasna:
👍 @MelissaAutumn In theory this only hits already existing `msgid`s so it _should_ actually pick them up right away!

EDIT: Looks great on stage, thanks!

MelissaAutumn:
> 👍 @MelissaAutumn In theory this only hits already existing `msgid`s so it _should_ actually pick them up right away!

True for most cases, but sometimes string extraction can be a little wacky. Will double check.

PR #853, opened 2025-06-16
Expose smart CTA "other" downloads to l10n

janbrasna:
See #851 followup to #852

Also adding smart slim CTA `"other"` downloads to gettext, terms already existing.


MelissaAutumn (approved):
Thank you!

PR #880, opened 2025-07-11
add new hires

malini:
checked a build locally and it's looking good


malini:
oh cool, good catch! will do

malini:
@MelissaAutumn thanks for the catch! updated the positions and verified locally with a build

MelissaAutumn:
Rad, thanks!

PR #897, opened 2025-08-05
Update index.html - add the participate URL with the locale in it

bogomil:
Update index.html - add the participate URL with the locale in it


MelissaAutumn (approved):
Thank you for the PR!

Issue #885, opened 2025-07-24
Thunderbird Pro Website


MelissaAutumn:
## Details

It's time to launch a new tbpro marketing / sign up page. Attached to this ticket are sub issues which is the bulk of the work split-up. 

This ticket is for all the general work on top of the sub-issues (so the bulk of the content.) Plan information and quotes have been blacked out because they're placeholder data. **Make sure this data is obviously placeholder data in the finished work.**

Additionally the copy is incoming (1-2 weeks I believe), so just fill it in with the lorem ipsum for now.

The screenshots

* [ ] Create a new buildable tb.pro site under build_site.py
* [ ] Create a new stylesheet base for tb.pro and integrate web guidelines from bolt.thunderbird.net
* [ ] Create base template w/ footer for tb.pro
* [ ] Create service page template
* [ ] Create tb.pro button component
* [ ] Create quote carousel component
* [ ] Create subscription plan component
* [ ] Create blamity bar component 
* [ ] Add the blamity bar component to www.thunderbird.net site

## Zeplin
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/6877e608631111e4347f7bd3

## Screenshots

![Image](https://github.com/user-attachments/assets/3a5fd157-f2f5-4261-8111-b437aaea7d7c)
![Image](https://github.com/user-attachments/assets/86551c1e-d87f-433b-a7da-1c908ec1eb00)

laurelterlesky:
@radishmouse Here are optimized images for the TB Pro Landing page.

Be sure to check the revised Zeplin page too: 
https://zpl.io/xnnDD7m

[img-optimized-TBProLanding.zip](https://github.com/user-attachments/files/22805950/img-optimized-TBProLanding.zip)

PR #900, opened 2025-08-18
Add new tbpro target to build-site.py

### LINKED ISSUE #886: Create a new buildable tb.pro site under build_site.py
## Details

We'll need to setup a new build target in build-site.py for tb.pro. This includes creating a folder under the sites directory. 

While we're working on this if there's any easy clean-ups we can do to the existing code we should get that done.

radishmouse:
Closes #886
(Part of #885 work)

* Reorganize `build-site.py` into functions
* Handle new `--tbpro` argument in `build-site.py`
* Add `build_tbpro` function to `builder.py`
* Update README.md with mention of `--tbpro` argument
* Add values to `settings.py` for tb.pro
* Add dummy `index.html` for tb.pro


build-site.py
```diff
@@ -1,108 +1,184 @@
...
+# Run builders based on command line args.
+site = None
+for arg, handler in build_handlers.items():
+    if getattr(args, arg):
+        # If a handler returns a value, it will be a `site` instance.
+        val = handler()
+        if val:
+            # Set this so the watcher can start, if requested.
+            site = val
+        break
```
MelissaAutumn: While the rest of the changes to build-site.py probably should happen at some point, this change in particular is kinda just odd? I see where you're coming from, but I think it's probably easier to just use an if or a match statement here.

build-site.py
```diff
@@ -1,108 +1,184 @@
...
+def prepare_updates_context():
+    """Prepare the context dictionary for the updates site build."""
+    default_channel = settings.DEFAULT_RELEASE_VERSION
+    version = helper.thunderbird_desktop.latest_version(default_channel)
+    beta_version = helper.thunderbird_desktop.latest_version('beta')
+
+    return {
         'current_year': date.today().year,
         'matomo_site_id': settings.MATOMO_SITE_IDS.get('utn'),
     }
```
MelissaAutumn: These context functions are only used once, I think we can merge them where they're used.

build-site.py
radishmouse: Just to make sure I'm clear, is one of your suggestions to replace the `if val: site = val` with this?

```
match val:
    case None:
        pass
    case _:
        site = val
```
(Or maybe just being more explicit with `if val is not None`?)

Apologies if it's super obvious and I'm just missing it!

MelissaAutumn: More so like moving it back to 
```
if args.startpage: 
  build_startpage()
elif args....etc....
```

radishmouse: Ah!
Sorry - I got thrown off by the quoted code.

Yep :+1: 
Can definitely go back to the if/elif for choosing which builder to run.

build-site.py
```diff
@@ -82,27 +126,46 @@
...
+    download_legal()
+elif args.tbpro:
+    site = build_tbpro()
+else:
+    site = build_main_website()
 
-if args.watch:
+# Set up file watcher if requested.
+if args.watch and site:
     builder.setup_observer(site, args.port)
```
MelissaAutumn, line 173: startpage and updates should also be observed if requested.

MelissaAutumn (changes requested):
Sorry one more thing and I'm almost in a place where I can test things :sweat_smile:

build-site.py
radishmouse: Good catch!
Updated.

radishmouse:
> Thanks you were missing two things, the _media symlink (not explained anywhere, sorry 😅) and the return vals for startpage and updates.
> 
> Looks like we're working again!

lol I just remembered the return vals as I closing the laptop for a break.
Thanks for fixing!

PR #918, opened 2025-09-18
Tbpro style and content updates 09-18

radishmouse:
A few tweaks:

* Remove `@md-max` breakpoint in favor of `@lg`
* Update copy for "Email Confidently" 
* Constrain main containers to `@lg` (1280px / 80rem)

### DIFF
--- a/assets/less/tbpro-style.less
+++ b/assets/less/tbpro-style.less
@@ -398,7 +398,6 @@ section#details {
         background-color: rgba(0, 0, 0, 0.2);
         margin-block: @space-36;
         padding-inline: @space-36;
-        outline: 1px solid blue;
 
         display: grid;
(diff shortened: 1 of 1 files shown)


assets/less/tbpro-style.less
```diff
@@ -1,5 +1,17 @@
...
+#masthead,
+#features,
+#reasons,
+#quotes,
+#details,
+#faq {
+        max-width: @lg;
+        margin-inline: auto;
+
+}
```
MelissaAutumn, line 13: Can you add a class to these sections, maybe titled `content-container`, or `content-section`? I like the ids (for anchor tagging urls) though just best to avoid using them when adding a single class to each will reduce this to like 4 lines.

PR #920, opened 2025-09-19
Create a lambda that bridges donor form responses to Zendesk. (Fixes #919)(#920)

### LINKED ISSUE #919: Create a bridge for the donor form to Zendesk's API

Sancus:


tools/form-bridge/lambda_function.py
```diff
@@ -0,0 +1,154 @@
...
+    'tfa_196': 'Help with something else'
+}
+
+def map_form_to_zendesk(form_data: Dict[str, str]) -> Dict[str, Any]:
+    """Map form fields to Zendesk ticket format"""
+    # Map basic fields
+    mapped_data = {
+        zendesk_field: form_data.get(form_field)
+        for form_field, zendesk_field in FIELD_MAPPINGS.items()
+        if form_field in form_data
```
MelissaAutumn, line 62: Does order matter how? I thought if would have to be before the for loop. :thinking:

tools/form-bridge/lambda_test.py
```diff
@@ -0,0 +1,189 @@
...
+        # Missing tfa_10 (email)
+    }
+
+    result = map_form_to_zendesk(form_data)
+
+    # Should use defaults for missing fields
+    assert result['ticket']['subject'] == 'Support Request'
+    assert result['ticket']['comment']['body'] == 'No comment provided'
+    assert result['ticket']['requester']['name'] == 'John Doe'
+    assert result['ticket']['requester']['email'] == ''
```
MelissaAutumn, line 189: Thanks for the tests! Could use some bad input or bad base64 encoding tests to round things off though!

tools/form-bridge/lambda_function.py
Sancus: This is a dict comprehension spread over multiple lines so the order is correct. But maybe it shouldn't be a comprehension if it's confusing to read...

tools/form-bridge/lambda_test.py
Sancus: Added tests, fixed a bug and improved the README a little.

tools/form-bridge/lambda_test.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+
```
MelissaAutumn: nit spacing :smile:

tools/form-bridge/lambda_test.py
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        result = lambda_handler(event, None)
+        assert result['statusCode'] == 200
+
+        call_args = mock_send.call_args[0][0]
+        assert call_args['ticket']['subject'] == 'Technical support for Thunderbird - No subject provided'
+        assert call_args['ticket']['comment']['body'] == 'No comment provided'
+        assert call_args['ticket']['requester']['name'] == 'No name provided'
+        assert call_args['ticket']['requester']['email'] == ''
```
MelissaAutumn: While this is valid, do we want to error out here? I would assume we want at least their email address otherwise the submission is just spam.

tools/form-bridge/lambda_test.py
```diff
@@ -0,0 +1,305 @@
...
+        assert response_body['error'] == 'Failed to submit form'
+
+def test_special_characters_in_form_data():
+    """Test handling of special characters, Unicode, and potential XSS attempts"""
+    form_data = {
+        'tfa_211': 'Test with émoji 🎉 and special chars',
+        'tfa_163': '<script>alert("XSS")</script> & HTML entities',
+        'tfa_1': 'José María Núñez',  # Unicode characters
+        'tfa_10': 'test+user@example.com'  # Plus sign in email
+    }
```
MelissaAutumn, line 320: Classic python 2 test conditions :smile:

MelissaAutumn (approved):
New tests look good. But it raised the question of whether we should actually allow empty string or null email addresses in submissions.

I'm not sure there's a valid reason to submit this form unless you want a reply back. I'll keep this approved though because we can punt that to another ticket.

tools/form-bridge/lambda_test.py
Sancus: So, right now, what will happen is we pass the empty email but Zendesk API will fail and so we'll return a 500 on that basis.

I guess it's reasonable to fail before we even call Zendesk here, though I don't really want to try to duplicate all the potential Zendesk error conditions.

But in this case I agree there can't be a legitimate request with an empty email.

PR #923, opened 2025-09-24
[DO NOT MERGE] Draft of design system buttons

radishmouse:
This is an initial attempt at implementing the [Buttons from the Bolt Design System](https://www.figma.com/design/1wGTkoIgRfyEbLeiW1XeIx/Design-System--%3E-Components?node-id=2035-915&m=dev).

It makes use of a few features from LESS:
- concrete classes `:extend` a handful of base classes
- concrete classes call mixin functions to add additional styles
- LESS variables used locally in rulesets to improve readability

Things to ignore:
- Colours have been redefined in the `:root` in buttons.less, just out of convenience while developing. They will be properly matched up with the definitions in colours.less if there are no issues with the approach taken in this PR.
- There is going to be some repetition of literal values (mainly colours), just because I didn't spend the time to go through and clean it up as much as I could have.
- In some places `px` units are used instead of `rem` units. This is was only done to save time.
- any changes other than what is in `buttons.less`

Would like feedback on the code in `buttons.less`, specifically:
- the way that the mixins are written and used
- the relationship between concrete classes and parent classes

### DIFF
--- a/assets/less/tbpro/buttons.less
+++ b/assets/less/tbpro/buttons.less
@@ -1,0 +1,432 @@
+:root {
+    --height: 2.75rem; // 44px
+    --height-small: 2.25rem; // 36px
+    --radius: 0.5rem; // 8px
+    --border-width: 1px;
+    --button-text-color: #1A202C;
+    --font-color: var(--text-icon-base);
+    --font-size: 0.875rem;
+    --critical-default: #991B1B;
+    --critical-hover: #7F1D1D;
+    --critical-pressed: #450A0A;
+    --primary-hover: #175FB6;
+    --primary-pressed: #19518F;
+    --surface-border: #E4E4E7;
+    --surface-base: #FEFFFF;
+    --surface-lower: #F7F7F7;
+    --surface-deep: #E4E4E7;
+    --text-icon-muted: #737584;
+
+}
+
+// ================================================================================
+// Concrete classes for each kind of button
+//
+.button-primary-filled-label:extend(.button-default) {
+    .primary-filled-mixin();
+    .primary-filled-focus-mixin();
+}
+
+.button-primary-outline-label:extend(.button-default) {
+    .primary-outline-mixin();
+}
+
+.button-primary-small-filled-label:extend(.button-small) {
+    .primary-filled-mixin();
+}
+
+.button-primary-small-outline-label:extend(.button-small) {
+    .primary-outline-mixin();
+}
+
+.button-brand-filled-label-icon:extend(.button-brand-default) {
+    .brand-filled-icon-mixin();
+    .brand-filled-icon-focus-mixin();
+}
+
+.button-brand-outline-label:extend(.button-brand-default) {
+    .brand-outline-mixin();
+    .brand-outline-focus-mixin();
+}
+
+.button-brand-small-outline-label:extend(.button-brand-small) {
+    .brand-outline-mixin();
+    .brand-outline-focus-mixin();
+}
+
+.button-destructive-filled-label:extend(.button-default) {
+    .destructive-filled-mixin();
+}
+
+.button-destructive-outline-label:extend(.button-default) {
+    .destructive-outline-mixin();
+}
+
+.button-destructive-small-filled-label:extend(.button-small) {
+    .destructive-filled-mixin();
+}
+
+
+// ================================================================================
+// "Parent classes" for concrete buttons
+//
+.button-default {
+    display: inline-flex;
+    justify-content: center;
+    align-items: center;
+    flex-shrink: 1;
+
+    height: var(--height);
+
+    margin: 0;
+    padding: 1.125rem 1rem;
+
+    box-shadow: 0 0.125rem 0.25rem 0 rgba(0, 0, 0, 0.12);
+    border: none;
+    border-radius: var(--radius);
+    cursor: pointer;
+
+    // TODO: font styles - family, height
+    font-size: var(--font-size);
+}
+
+.button-small:extend(.button-default) {
+    height: var(--height-small);
+    padding: 0.375rem 0.75rem;
+}
+
+.button-brand-default:extend(.button-default) {
+    padding: 1rem 1.5rem;
+
+    font-size:  0.8125rem;
+    color: var(--text-icon-base, #1A202C);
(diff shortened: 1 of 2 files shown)


assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+    --critical-pressed: #450A0A;
+    --primary-hover: #175FB6;
+    --primary-pressed: #19518F;
+    --surface-border: #E4E4E7;
+    --surface-base: #FEFFFF;
+    --surface-lower: #F7F7F7;
+    --surface-deep: #E4E4E7;
+    --text-icon-muted: #737584;
+
+}
```
radishmouse, line 20: These are only some of the values used throughout this file.
It was partially just because it was easier to paste from figma than it was to find the corresponding definitions in `base.less`

assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+
+// ================================================================================
+// Concrete classes for each kind of button
+//
+.button-primary-filled-label:extend(.button-default) {
+    .primary-filled-mixin();
+    .primary-filled-focus-mixin();
+}
```
radishmouse, line 28: One way to read these selectors:
> A primary-filled-label button is a default button, with "primary filled" and "primary filled focus" styles.

assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+    .primary-outline-mixin();
+}
+
+.button-primary-small-filled-label:extend(.button-small) {
+    .primary-filled-mixin();
+}
+
+.button-primary-small-outline-label:extend(.button-small) {
+    .primary-outline-mixin();
+}
```
radishmouse, line 40: The `*-focus-mixin()` is separate because the design does not define focus styles for every kind of button.

assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+
+.button-destructive-small-filled-label:extend(.button-small) {
+    .destructive-filled-mixin();
+}
+
+
+// ================================================================================
+// "Parent classes" for concrete buttons
+//
```
radishmouse, line 72: These are styles that are common to the main categories of buttons.
I thought of them like base or parent classes that the more specific ones inherit from.

They are not used directly.

assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+
+.button-brand-small:extend(.button-brand-default, .button-small) {
+    font-size:  0.6875rem;
+    padding: 0.875rem 1rem;
+}
+
+
+// ================================================================================
+// Button variant mixins
+//
```
radishmouse, line 123: The style mixins apply backgrounds, borders, and some font styles.
This is where the different button states are defined.

assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+        font-weight: 400;
+        line-height: normal;
+
+    }
+}
+
+.primary-filled-focus-mixin() {
+    // Using :focus-visible so that it's only for keyboard focus.
+    // Regular :focus overrides the :active psuedo-class.
+    &:focus-visible {
```
radishmouse, line 157: Note the reason why `:focus-visible` is used.

assets/less/tbpro/buttons.less
```diff
@@ -0,0 +1,432 @@
...
+
+// ================================================================================
+// Mixin classes for filled and outline buttons.
+//
+.filled(@background-color, @font-color: var(--font-color)) {
+    background: @background-color;
+    color: @font-color;
+}
+
+.outline(@border-color, @background-color: transparent, @border-width: var(--border-width)) {
```
radishmouse, line 304: The three `.outline()` mixins apply a weird hack that allows `linear-gradient()` to be used for borders. (They're not real borders - they're `::before` and `::after` with a hole in the middle.

assets/less/tbpro/buttons.less
davinotdavid: This is an interesting point. Unsure the best way to enforce this (I've used [Tokens Studio](https://www.figma.com/community/plugin/843461159747178978/tokens-studio-for-figma) as JSON files a long time ago but wasn't the best experience maybe now it's better) but I believe we should push for "ubiquitous language" for variables used cross design <-> dev to alleviate this.

assets/less/tbpro/buttons.less
davinotdavid: IMHO, it would be nice to point that these shouldn't be used directly in the comment section just in case!

assets/less/tbpro/buttons.less
MelissaAutumn: Our hope is to get bolt design systems setup fully, and look into design tokens (sort of what davi has suggested there.) From there, us in the services team will have a magical token to css/less/whatever codegen CI or tool setup. Shouldn't be too hard to find an existing tool, or to build one that matches our schema. 

You can see a small sample of the tokens here: https://bolt.thunderbird.net/8b179dbfd/p/824dfe-design-tokens (including json export) but they're not the ones for web.

Unfortunately everyone has been busy so this hasn't really progressed. So thanks for adding this for now, hopefully we can slot in the automated css/less in the future.

assets/less/tbpro-style.less
```diff
@@ -1,5 +1,17 @@
 @import "tbpro/base.less";
 
+// Constrain all main blocks to the largest breakpoint.
+#masthead,
+#features,
+#reasons,
+#quotes,
+#details,
+#faq {
```
MelissaAutumn, line 9: Ties into #924 :smile:

assets/less/new/layout.less
```diff
@@ -457,7 +457,7 @@ section {
 
 .carousel > picture:last-child {
   opacity: 0;
-  animation-duration: 30s;
+  animation-duration: .3s;
```
MelissaAutumn, line 460: Left over from debugging?

radishmouse: Haha yep. Debug droppings!

I branched off of some other WIP code. Really only meant to focus on buttons.less in this PR.

radishmouse:
Thanks for the feedback!
Closing this one to reduce GitHub PR noise-

PR #927, opened 2025-10-03
added the requested link

GTgyani206:


hellsworth (changes requested):
I think We want to use `class="strong"` and curly braces like in other places of this index.html where url's are defined. for example see [Line 41](https://github.com/thunderbird/thunderbird-website/blob/master/sites/www.thunderbird.net/about/index.html#L41)

GTgyani206:
@hellsworth hi, sorry I should have noticed the pattern before... Well, I have made the changes now, but I just wanted to confirm if this is the correct way to add the link:  ```<a class="strong" href="{{ url( 'council.thunderbird.net') }}">{{ _('Learn more about the Thunderbird Council') }}</a>```. If this looks good, I’ll go ahead and push the commits.
Also this is the first time I've come across this type of anchor tag, so can you please educate me a little on why are we using this type and how is it different from the traditional one....

hellsworth:
> If this looks good, I’ll go ahead and push the commits.

Looks good to me

>can you please educate me a little on why are we using this type and how is it different from the traditional one

Sorry, I'm not a webdev by any means so I don't know the history of why this syntax is being used. This is more of a question for @Sancus or @MelissaAutumn

GTgyani206:
> Sorry, I'm not a webdev by any means so I don't know the history of why this syntax is being used.

No worries ☺️... Thanks though!!

sites/www.thunderbird.net/about/index.html
```diff
@@ -122,6 +122,7 @@ <h2 aria-label="{{ _('Governed by Contributors') }}">{{ _('Governed by <span cla
             this council is elected by contributors and acts as their voice in determining Thunderbird's strategic
             direction and goals. The Council oversees funds, approves budgets, and shapes the product roadmap,
             ensuring Thunderbird remains true to its core values.
+	   <a class="strong"  href="{{ url('council.thunderbird.net') }}">{{ _('Learn more about the Thunderbird Council') }}</a>
```
Sancus: Thanks for the PR. So first of all, this link should not be part of the `<p>` tag. It should be below the closing `</p>`. Otherwise it doesn't match the other links of this type on the page.

Secondly, 

The `url()` function is Python and pulls the link from `URL_MAPPINGS` in https://github.com/thunderbird/thunderbird-website/blob/master/settings.py#L167 -- it exists as a shortcut for longer links and to make it simple to change a link when necessary. You would need to add the link there for this to work.

Given this is a single link not likely to be reused anywhere else anytime soon, you don't need to use it, `https://council.thunderbird.net` is fine.

GTgyani206:
Hey @Sancus  can you please review the commit

Sancus:
You'll want to swap `{{ url('council.thunderbird.net') }}` for just `https://council.thunderbird.net` since you're not adding the url setting to settings.py

PR #943, opened 2025-10-20
Create shared tb/tbpro components

radishmouse:
Closes #893
Closes #930 

* Creates and styles calamity bar (nav bar that switches between tb website and tb.pro website)
* Create shared version of footer
* Use shared calamity bar and footer on tb.pro
* Update build scripts to include and watch (optional) shared template paths
* Populate footer with URLs from `settings.py`

PR #944, opened 2025-10-21
Restore footer on thunderbird.net

radishmouse:
Addressing issue raised here: https://github.com/thunderbird/thunderbird-website/pull/943#issuecomment-3424667817

This PR restores the original footer for www.thunderbird.net.


Sancus:
Thanks!

PR #960, opened 2025-10-28
Add build+test CI pipeline

### LINKED ISSUE #863: Run pytest per PR via github actions
We should run pytest for every PR to ensure functionality doesn't break.

janbrasna:
This GitHub Actions workflow runs a matrix of a few select Python and Node versions on several platforms/archs to verify sites build and tests run for PRs and master pushes; with visibility into deprecations and incompatibilities for future versions.

Resolves #863

👓 demo/preview: [`@janbrasna/thunderbird-website` /actions/runs/18867825525](https://github.com/janbrasna/thunderbird-website/actions/runs/18867825525/job/53838925997)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,8 +44,8 @@ jobs:
       run: |
         npm install -g less
         pip install -r requirements-dev.txt
-        git clone https://github.com/thunderbird/thunderbird-notes.git libs/thunderbird_notes
-        git clone -b production https://github.com/mozilla-releng/product-details.git libs/product-details
+        git clone --depth 1 https://github.com/thunderbird/thunderbird-notes.git libs/thunderbird_notes
+        git clone --depth 1 -b production https://github.com/mozilla-releng/product-details.git libs/product-details
     - name: Build site (just on darwin)
       if: runner.os == 'macOS'
       run: |


.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Build tbpro (just on darwin)
+      if: runner.os == 'macOS'
+      run: |
+        python build-site.py --debug --enus --tbpro
+  # - name: Build calendars (disabled until new provider)
+  #   run: |
+  #     python build-site.py --debug --enus --buildcalendar
+    - name: Run pytest
+      run: |
+        python -m pytest tests
```
janbrasna: I'm specifically restricting this to `./tests` as pytest otherwise collects from places that don't have their dependencies installed, i.e. enumerated in any of the requirements files. (A separate issue in itself.)

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+          os: windows-2025
+        - python: "3.13"
+          node: 22
+          os: ubuntu-24.04
+        - python: "3.14"
+          node: 24
+          os: macos-26
+        - python: "3.15-dev"
+          node: 25
+          os: ubuntu-latest
```
janbrasna: so why the platforms etc.?
_(was originally trying to find oldest versions; had to use non-latest runners)_. see issue
requirements, cached deps/stack, speed

MelissaAutumn:
This is rad, thanks! Could we switch the build-site to run off of ubuntu, and add a locale build? 

Locale compile fails are generally the biggest problem we have here, so it'd be nice to catch those during dev too! And the compile script (or maybe it's extract) uses gnuutils specific arguments that don't work on mac. (Also are servers are on centos iirc, so it's a little closer to irl)

```
git clone https://github.com/thunderbird/thunderbird.net-l10n.git libs/locale
l10n_tools/compile.sh
```

janbrasna:
Yes, I'll try to add Ubuntu at least, just currently facing a weird `system_requirements` [parsing issue](https://github.com/janbrasna/thunderbird-website/actions/runs/18854269574/job/53798128459#step:6:34) that for some reason does not exhibit on darwin, either for some toolcache differences, or os-level filesystem access specifics — so I at least started with one platform running it before discovering the blocker to add more.

I know there are currently some "fatal errors" in the gettext scripts from not having `trans trim` in all the places, but I guess I could inspect the return codes first to see if I can tell apart the long term failures from just real l10n crashes;) ~ +IIUC the only gnu compat wrinkle in the l10n scripts is the `find --printf="%h\n"` logging/verbosity for processed files/dirs that I can just edit out running the find directly in the CI (or maybe I can look into making the actual script more compatible with some `exec dirname` etc. instead?)

janbrasna:
Yeah the whitespace errors return with [codes](https://github.com/janbrasna/thunderbird-website/actions/runs/18885145837/job/53898435051#step:6:136) really so I can silence these; but then it won't catch any real issues, and be just a log… (NB: `msgfmt` is preinstalled on darwin, but not part of `gettext-base` on ubuntu, so it's additional 2min+ to apt-install it in flight each run:/…)

I will try to fight the yaml parsing on platforms for a bit, but it either doesn't see the key, or [bails](https://github.com/janbrasna/thunderbird-website/actions/runs/18885145837/job/53898435016#step:9:49) on encoding. 🤷

MelissaAutumn:
Ah I see, that's pretty odd. Seems like a problem in thunderbird notes. I see https://github.com/thunderbird/thunderbird-notes/issues/70 has already been made, could you create a bug ticket for the system_requirements problem when you have a chance? 

Thank you, it's great to see this has already found some blocking bugs for contributors :smile:

PR #1030, opened 2025-11-24
Draft of styles for Dec 25 Appeal, versions 1a and 1b

radishmouse:
Note: this is not the final version

* Styles for 1a/1b
* Includes final copy
* Does not include illustration


radishmouse:
@MelissaAutumn I'm going to go ahead and merge so it goes to stage.
I'm not sure how complete it needs to be before we extract strings for translation, but I imagine we can't really wait too much longer.

PR #1033, opened 2025-11-25
Fix tab navigation for staging environment

radishmouse:
* Highlight "Thunderbird" tab for either `stage.` or `www-stage.`
* Remove hard-coded `"en-US"` from URLs

### DIFF
--- a/assets/less/common/tab-navigation-bar.less
+++ b/assets/less/common/tab-navigation-bar.less
@@ -43,6 +43,7 @@ body.tbpro #tab-navigation-bar {
         );
         align-items: center;
         color: @color-nav;
+        cursor: pointer;
         display: flex;
         padding-inline: 20px;
         text-decoration: none;
@@ -52,6 +53,9 @@ body.tbpro #tab-navigation-bar {
     a:hover {
         background-color: @bg-active;
         color: @color-active;
+    }
+
+    a.active {
         cursor: default;
     }
 }


sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -8,8 +8,9 @@
 <script>
   const currentHost = window.location.hostname;
   const isStage = currentHost === 'www-stage.thunderbird.net' ||
-                  currentHost === 'stage.tb.pro' ||
-                  currentHost === 'localhost';
+        currentHost === 'stage.thunderbird.net' ||
+        currentHost === 'stage.tb.pro' ||
+        currentHost === 'localhost';
```
MelissaAutumn: At some point @Sancus we should really move to hosting the current environment in an environment variable or something. 

@radishmouse can we shove these values into settings.py like:

```python

... around CANONICAL_URL ...

CANONICAL_STAGE_URL = 'https://www-stage.thunderbird.net'
STAGE_HOSTS = [CANONICAL_STAGE_URL, 'stage.thunderbird.net', 'stage.tb.pro', 'localhost']

```

Then you should be able to just use them like

```
const stageHosts = {{ settings.STAGE_HOSTS }};
const isStage = stageHosts.indexOf(currentHost) !== -1;
```

In this case the variable is a python list so this should work as the syntax is pretty 1:1. But if STAGE_HOSTS were a tuple or other data type you might want to do: 

```
const stageHosts = {{ settings.STAGE_HOSTS|tojson|safe }};
```

Which will use the [tojson filter](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.tojson) to convert the variable to json. The safe filter just ensures the string isn't sanitized, it might also be the one decoding the json that part I'm not sure. 🤔

radishmouse: Can do! Commit forthcoming

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -6,10 +6,9 @@
 </div>
 
 <script>
+  const stageHosts = {{ settings.STAGE_HOSTS|tojson|safe }};
   const currentHost = window.location.hostname;
-  const isStage = currentHost === 'www-stage.thunderbird.net' ||
-                  currentHost === 'stage.tb.pro' ||
-                  currentHost === 'localhost';
+  const isStage = stageHosts.includes(currentHost);
```
MelissaAutumn, line 11: I'm so stuck in IE6 compatiable javascript from 2013, I forgot includes exists.

sites/common/includes/base/tab-navigation-bar.html
MelissaAutumn: Can we use {{ settings.CANONICAL_URL }} here?

MelissaAutumn: We might have to make a CANONICAL_TBPRO_URL

MelissaAutumn: Can we add `data-link-id="thunderbird"` or something similar?

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -18,10 +17,10 @@
     // Determine which link this is and update.
     if (linkHost.includes('thunderbird.net')) {
```
MelissaAutumn: With the `data-link-id` you can query link.dataset['link-id'] (or something to that effect) instead of relying on a partial url match.

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -6,10 +6,9 @@
...
 <script>
+  const stageHosts = {{ settings.STAGE_HOSTS|tojson|safe }};
   const currentHost = window.location.hostname;
-  const isStage = currentHost === 'www-stage.thunderbird.net' ||
-                  currentHost === 'stage.tb.pro' ||
-                  currentHost === 'localhost';
+  const isStage = stageHosts.includes(currentHost);
 
   // Update href values based on environment
   document.querySelectorAll('#tab-navigation-bar a').forEach(link => {
```
MelissaAutumn: Can we only run this if isStage is true?

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -18,10 +17,10 @@
     // Determine which link this is and update.
     if (linkHost.includes('thunderbird.net')) {
       // Thunderbird link
-      link.href = isStage ? 'https://www-stage.thunderbird.net/en-US/' : 'https://www.thunderbird.net/en-US/';
+      link.href = isStage ? 'https://www-stage.thunderbird.net/' : 'https://www.thunderbird.net/';
```
MelissaAutumn: These should use the canonical urls from settings, likewise for tb.pro.

radishmouse:
@MelissaAutumn I started with your suggestions and simplified a bit further.
Hopefully readable and sensible?

settings.py
```diff
@@ -104,7 +104,16 @@
...
 
 CANONICAL_URL = 'https://www.thunderbird.net'
+CANONICAL_STAGE_URL = 'https://www-stage.thunderbird.net'
 CANONICAL_UPDATES_URL = 'https://updates.thunderbird.net'
+CANONICAL_TBPRO_URL = 'https://www.tb.pro'
+CANONICAL_TBPRO_STAGE_URL = 'https://stage.tb.pro'
+
+STAGE_HOSTS = [
+    CANONICAL_STAGE_URL.removeprefix('https://'),
+    CANONICAL_TBPRO_STAGE_URL.removeprefix('https://'),
```
radishmouse, line 114: `.removeprefix()` requires Python 3.9
If our build pipeline uses something older, I can adjust these lines.

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -1,40 +1,33 @@
 <div id="tab-navigation-bar">
   <div class="tab-navigation-links">
-    <a href="https://www.thunderbird.net">Thunderbird</a>
-    <a href="https://www.tb.pro" class="active">Thunderbird Pro</a>
+    <a href="{{ settings.CANONICAL_URL }}" data-stage-url="{{ settings.CANONICAL_STAGE_URL }}">Thunderbird</a>
+    <a href="{{ settings.CANONICAL_TBPRO_URL }}" class="active" data-stage-url="{{ settings.CANONICAL_TBPRO_STAGE_URL}}">Thunderbird Pro</a>
```
radishmouse, line 4: Decided to do away with looping and matching. Realized that if we tag each link with an ID, might as well just tag each one with the eventual URL to swap.

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -1,39 +1,33 @@
...
+  }
 
   // Set .active class accordingly.
   document.querySelectorAll('#tab-navigation-bar a').forEach(link => {
-    const linkHost = new URL(link.href).hostname;
+    const linkHost = new URL(link.href).hostname.replace('www', '');
+    const doesMatchCurrent = linkHost === currentHost;
 
-    if (linkHost.replace('www.', '') === currentHost.replace('www.', '')) {
+    if (doesMatchCurrent) {
```
MelissaAutumn: nit: You can move that var to here since it's only used once

sites/common/includes/base/tab-navigation-bar.html
```diff
@@ -1,39 +1,33 @@
...
       link.classList.add('active');
+      link.style.cursor = 'default';
```
MelissaAutumn: You should add this to the active class instead of styling it here.

settings.py
MelissaAutumn: We should at least be using Python 3.11 especially since Python 3.9 is EOL'd https://devguide.python.org/versions/

MelissaAutumn (approved):
Looks good after these last few fixes

PR #1043, opened 2025-12-03
Fix height difference of tab nav bar across tb.net and tb.pro

radishmouse:
Fixes the 4px difference in height of tab-navigation-bar between www and tb.pro

Details:
* tb.pro styling used a negative margin to match the design
* The negative margin caused 4px of the tab-navigation-bar to be hidden
* Fixed by adjusting the padding and removing the negative margin


MelissaAutumn (approved):
Looks good! I see we have some font size animations happening on tbnet, but that was there before, and I think caused by the font loading in.

PR #1046, opened 2025-12-04
Add container and ECS Fargate deploying with Pulumi.

Sancus:
This adds a GitHub workflow, Pulumi code, and a Docker container that hosts this website and all the other ones hosted by our Apache setup.

mx and broker don't work yet but that's not super important.

This also puts the web servers onto a modern CentOS with Python3.

https://www-stage.thunderbird.net is currently running on this.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,9 +13,9 @@ on:
   repository_dispatch:
     types: [stage, prod]
   # Allow testing from feature branches
-  push:
-    branches:
-      - fargate
+  # push:
+  #   branches:
+  #    - fargate
 
 permissions:
   id-token: write


.github/workflows/container-test.yml
```diff
@@ -0,0 +1,131 @@
...
+          load: true
+          tags: thunderbird-web:test
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Start container
+        run: |
+          docker run -d --name tb-web-test -p 8080:80 thunderbird-web:test
+          # Wait for Apache to start
+          sleep 5
```
MelissaAutumn: We should cut a ticket to switch this to a healthcheck wait loop in the future.

.github/workflows/container-test.yml
```diff
@@ -0,0 +1,131 @@
...
+          echo "$response"
+          echo "$response" | grep -qE "(200 OK|302 Found)"
+
+      - name: Show container logs on failure
+        if: failure()
+        run: |
+          echo "=== Container logs ==="
+          docker logs tb-web-test
+          echo "=== Apache error log ==="
+          docker exec tb-web-test cat /var/log/httpd/error_log || true
```
MelissaAutumn, line 129: It'd be nice if these were uploaded, but this works too.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+  group: deploy-${{ github.event.inputs.environment || 'stage' }}
+  cancel-in-progress: true
+
+env:
+  AWS_REGION: us-west-2
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment: ${{ github.event.inputs.environment || (github.event.action == 'deploy-prod' && 'prod' || 'stage') }}
```
MelissaAutumn: Since stage and prod are already specified in the environment as a choice of either what's up with the ors?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+        id: env
+        run: |
+          if [ -n "${{ github.event.inputs.environment }}" ]; then
+            ENV="${{ github.event.inputs.environment }}"
+          elif [ "${{ github.event.action }}" = "deploy-prod" ]; then
+            ENV="prod"
+          else
+            ENV="stage"
+          fi
+          echo "environment=$ENV" >> $GITHUB_OUTPUT
```
MelissaAutumn: Can you not use ENV={{ inputs.environment }}?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+    branches:
+      - fargate
+
+permissions:
+  id-token: write
+  contents: read
+
+# Cancel in-progress deployments when a new one starts
+concurrency:
+  group: deploy-${{ github.event.inputs.environment || 'stage' }}
```
MelissaAutumn, line 26: environment is a required choice, is there any situation where this will be null?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+          fi
+          echo "environment=$ENV" >> $GITHUB_OUTPUT
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::768512802988:role/thunderbird-website-deploy
```
MelissaAutumn, line 46: Might want to punt this to an env at some point. Fine for now.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+            -t $ECR_REGISTRY/thunderbird-website-$BUILD_ENV:latest \
+            .
+          docker push $IMAGE_URI
+          docker push $ECR_REGISTRY/thunderbird-website-$BUILD_ENV:latest
+          echo "image_uri=$IMAGE_URI" >> $GITHUB_OUTPUT
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
```
MelissaAutumn, line 80: modern python hypee

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,130 @@
...
+          IMAGE_URI="$ECR_REGISTRY/thunderbird-website-$BUILD_ENV:$IMAGE_TAG"
+          docker build \
+            -f docker/Dockerfile \
+            --build-arg BUILD_ENV=$BUILD_ENV \
+            -t $IMAGE_URI \
+            -t $ECR_REGISTRY/thunderbird-website-$BUILD_ENV:latest \
+            .
+          docker push $IMAGE_URI
+          docker push $ECR_REGISTRY/thunderbird-website-$BUILD_ENV:latest
+          echo "image_uri=$IMAGE_URI" >> $GITHUB_OUTPUT
```
MelissaAutumn, line 75: Oh nice, we can start using BUILD_ENV on the website and move some client-side hacks to build-time.

docker/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+# For local builds: build the websites from source using a venv
+RUN if [ "$BUILD_ENV" = "local" ]; then \
+      echo "BUILD_ENV=local - building websites from source..." \
+      && cd /build \
+      && mkdir -p libs \
+      && git clone --depth 1 https://github.com/thunderbird/thunderbird.net-l10n.git libs/locale \
+      && git clone --depth 1 -b production https://github.com/mozilla-releng/product-details.git libs/product-details \
+      && git clone --depth 1 https://github.com/thunderbird/thunderbird-notes.git libs/thunderbird_notes \
+      && npm install -g less \
+      && python3.11 -m venv /build/venv \
```
MelissaAutumn, line 69: I think we can just use python3 here.

docker/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+      && rm -rf /var/www/html/start/.git \
+      && echo "Done cloning tb-website-builds"; \
+    fi
+
+# Set permissions
+RUN chown -R apache:apache /var/www/html /var/www/services /var/www/autoconfig.momo /var/log/httpd/autoconfig
+RUN rm -f /etc/httpd/conf.d/welcome.conf /etc/httpd/conf.d/autoindex.conf
+
+EXPOSE 80
+HEALTHCHECK --interval=30s --timeout=10s --start-period=10s --retries=3 \
```
MelissaAutumn, line 103: Do we want to hit the wsgi for the health check instead?

docker/Dockerfile
Sancus: Unfortunately required because system Python is 3.9 and it doesn't get replaced when you install a different one.

.github/workflows/deploy.yml
Sancus: This is a bit of a mess since repository_dispatch doesn't support the inputs type so you need to use `types` instead, but there's no reason you can't use the same strings for each and just make sure environment selection accounts for the two different sources.

I've cleaned it up some.

.github/workflows/deploy.yml
Sancus: It is, but there are some cases where it can be null. If you allow pushes directly from a branch(like the one set up currently for `fargate`), environment is not set, so I'm going to leave it in just as a safety.

docker/Dockerfile
Sancus: Yep and done.

.github/workflows/deploy.yml
Sancus: This whole section was not really needed and has been removed by simplifying the naming of the event types.

.github/workflows/container-test.yml
Sancus: Fixed this directly, it will health check every second for 30 seconds and then fail.

PR #1053, opened 2025-12-08
Ensure Apache uses https for rewrites when behind ALB. (Fixes #1052)

### LINKED ISSUE #1052: Apache incorrectly generates redirects to HTTP (not HTTPS)
https://bugzilla.mozilla.org/show_bug.cgi?id=2004389

This was probably caused by changing our load balancer scheme so SSL is terminated at the load balancer. Apache is generating http redirects.

You cannot actually pull the file over http, you get immediately redirected back to https, but this shouldn't happen in the first place.

Fix is to make Apache aware that the protocol is actually https.

Sancus:


Sancus:
> Looks good as far as the docs say. I don't really have a setup to test it though.

Refactored the tests to all run through pytest and simplify, and also added one to prove the http redirect issue is fixed. Which uncovered a similar issue in `wsgi.py` so I fixed that as well.

MelissaAutumn:
> > Looks good as far as the docs say. I don't really have a setup to test it though.
> 
> Refactored the tests to all run through pytest and simplify, and also added one to prove the http redirect issue is fixed. Which uncovered a similar issue in `wsgi.py` so I fixed that as well.

Hey, shoo! Go enjoy your holidays. If you're bored, go start a hobby project :smile: 

But thanks for the updated tests!

Sancus:
> > > Looks good as far as the docs say. I don't really have a setup to test it though.
> > 
> > 
> > Refactored the tests to all run through pytest and simplify, and also added one to prove the http redirect issue is fixed. Which uncovered a similar issue in `wsgi.py` so I fixed that as well.
> 
> Hey, shoo! Go enjoy your holidays. If you're bored, go start a hobby project 😄
> 
> But thanks for the updated tests!

Don't worry I've already stated working on https://makerworld.com/en/models/629788-modular-filament-drawer-system 😆

No more PRs from me, I was just annoyed by this bug's existence.

Issue #1062, opened 2025-12-13
The waitlist flow incorrectly redirected to thunderbird.net


kellymcs01:
The waitlist flow incorrectly redirected to thunderbird.net, which needs to be fixed by returning users to thunderbird.pro.

inattee:
@radishmouse using this for the waitlist confirmation page since it's essentially the same issue. 

Can we change the waitlist confirmation page to keep the user on the same page as the form but swap it with a message that says:

"Thank you for joining the waitlist for Thunderbird Pro Early Bird. We'll be in touch soon!"

<img width="1303" height="414" alt="Image" src="https://github.com/user-attachments/assets/c3a56046-ff32-4c42-b46c-8cfcc80a468b" />

radishmouse:
Closed per discussion in [this PR](https://github.com/thunderbird/thunderbird-website/pull/1102#issuecomment-3787222184)

tl;dr:
- MC does not officially let you do client-side form submissions.
- Workarounds are too hacky and/or somewhat not-secure.
- We'll be transitioning to another solution in the future.

PR #1102, opened 2026-01-20
Show custom confirmation when user joins tb.pro wait list

radishmouse:
Shows custom success/error message when user submits wait list form.

image

* Bundle jQuery for submitting JSONP, necessary for background form submission to Mailchimp
* Include new JS bundle on wait list page
* Modifies wait list page to submit form in background
* On success, hides form and welcome message, shows success message, as mentioned in https://github.com/thunderbird/thunderbird-website/issues/1062#issuecomment-3724901638
* On error, shows generic error message

### DIFF
--- a/sites/tb.pro/waitlist/index.html
+++ b/sites/tb.pro/waitlist/index.html
@@ -85,7 +85,7 @@ document.addEventListener('DOMContentLoaded', () => {
   const messageDiv = document.getElementById('waitlist-message');
   const welcomeDiv = document.getElementById('waitlist-welcome');
 
-  form.addEventListener('submit', (event) => {
+  function handleSubmit (event) => {
     event.preventDefault();
 
     // Clear out previous messages.
@@ -115,10 +115,17 @@ document.addEventListener('DOMContentLoaded', () => {
       },
       error: (err) => {
 
+        // On network error, fall back to regular submission.
+        form.removeEventListener('submit', handleSubmit);
+        form.action.replace('/post-json?', '/post?');
+        $(form).submit();
+
         messageDiv.innerHTML = "{{ wait_list_error_message }}";
         messageDiv.classList.remove('hidden');
       }
-    });
+    }
+
+  form.addEventListener('submit', handleSubmit);
   });
 });
 


radishmouse:
@inattee using the very generic message if something goes wrong: "An error occurred. Please try again."
Is that OK, or would you prefer something else?

sites/tb.pro/waitlist/index.html
```diff
@@ -65,7 +65,58 @@ <h2 id="form-header">{{ _('Join the Waitlist') }}</h2>
...
+    event.preventDefault();
+
+    // Clear out previous messages.
+    messageDiv.classList.add('hidden');
+    messageDiv.textContent = '';
+
+    // Modify URL so we can send JSONP to MC.
+    const actionUrl = form.action.replace('/post?', '/post-json?');
+
+    // Using jQuery, since fetch does not let us do JSONP.
```
MelissaAutumn, line 98: 🙀

sites/tb.pro/waitlist/index.html
```diff
@@ -65,7 +65,58 @@ <h2 id="form-header">{{ _('Join the Waitlist') }}</h2>
...
+    $.ajax({
+      url: actionUrl,
+      type: 'GET',
+      data: $(form).serialize(),
+      dataType: 'jsonp',
+      jsonp: 'c', // Mailchimp specifically requires 'c' as the callback param.
+      success: (data) => {
+        if (data.result === 'success') {
+          form.classList.add('hidden');
+          messageDiv.textContent = "{{ _('Thank you for joining the waitlist for Thunderbird Pro Early Bird. We\'ll be in touch soon!') }}";
```
MelissaAutumn: I don't think this will work because jinja templates are rendered when the site is bui...oh nevermind this will indeed work.

It might be less confusing if we assign these to jinja variables though at the top of the template like 

```
{%
set waiting_list_success_message = _(' ... ')
set waiting_list_error_message = _(' ... ') 
%} 

then

```
messageDiv.textContent = "{{ waiting_list_success_message }}"
```

I'd just double check if they extract correctly.

File: settings.py [hunk 3]
```diff
@@ -421,6 +421,10 @@
 }
 
 TBPRO_JS = {
+    'common-bundle': [
+        # For submitting wait list form as JSONP
+        'js/common/jquery-3.6.0.min.js'
+    ]
```
MelissaAutumn, line 427: I'd maybe name this bundle waiting-list-bundle since it'll only be used on the waiting list page.

sites/tb.pro/waitlist/index.html
```diff
@@ -65,7 +65,58 @@ <h2 id="form-header">{{ _('Join the Waitlist') }}</h2>
...
+
+  form.addEventListener('submit', (event) => {
+    event.preventDefault();
+
+    // Clear out previous messages.
+    messageDiv.classList.add('hidden');
+    messageDiv.textContent = '';
+
+    // Modify URL so we can send JSONP to MC.
+    const actionUrl = form.action.replace('/post?', '/post-json?');
```
MelissaAutumn, line 96: Why are we posting json instead of just regular post?

radishmouse: As far as I've been able to find, MC won't let folks do client-side submissions (via fetch or otherwise).
This implements a well-known workaround to do "fake-ajax" instead of going to the MC confirmation page.

This workaround (which is known as "JSONP", but I have no idea why, because there's no JSON), involves three steps:
1. the client sending a request to the server that includes the name of a JavaScript function
2. the server responding with a script tag that includes a call to that JS function
3. the client adding the script tag to the page (which automatically runs the function from step 1)

If and when the one-off function runs, the page knows that the "success" state has been reached.

Here's how this PR implements the JSONP workaround:
1. Use jQuery to dynamically create a uniquely named, one-off JavaScript function and add it to the page so it can be run later.
2. convert the form data to querystring params 
3. include the name of the one-off JS function in the querystring params.
4. send a GET request to the `/post-json` endpoint with all the qs params
5. receive a response from MC, which contains a script tag (that calls our uniquely named function)
6. celebrate because it worked and that was all very convoluted

At the `/post-json` endpoint, MC receives the data and adds the user to our mailing list as if though they submitted the form in a normal way.

And of course, this is all possible because MC provides the `/post-json` endpoint that works with this technique.

Et voila! Client side form submission without:
- API keys leaking on the page
- MC having to provide CORS headers (which they don't want to do for whatever reason)

sites/tb.pro/waitlist/index.html
radishmouse: haha. Just wait till you read https://github.com/thunderbird/thunderbird-website/pull/1102#discussion_r2710258746

sites/tb.pro/waitlist/index.html
radishmouse: PS: I didn't make this up: https://en.wikipedia.org/wiki/JSONP

MelissaAutumn: Could we reduce the complexity, but just pre-filling the form and having the user click on "Click on this button to confirm your wait list addition" (Or something with better copy)?

radishmouse: Just to make sure I understand, instead of doing client-side form submission:
- change the copy so the user knows they're going to MC
- leave the form as-is (so that it goes to MC)

MelissaAutumn: Yeah leave the form as-is, have it prefilled with the information we send and change the copy (with @inattee's approval) to "Join the waitlist" instead of Submit.

MelissaAutumn: Turns out I read this issue incorrectly. Whoops! This is fine.

sites/tb.pro/waitlist/index.html
```diff
@@ -5,6 +5,10 @@
 {% extends "includes/base/page.html" %}
 {% from 'includes/macros/right-arrow-submit-button.html' import right_arrow_submit_button %}
 
+{% set wait_list_success_message = _('Thank you for joining the waitlist for Thunderbird Pro Early Bird. We\'ll be in touch soon!') %}
+{% set wait_list_error_message =_('An error occurred. Please try again.') %}
```
MelissaAutumn, line 9: Mailchimp submission seems to be broken thanks to Firefox's enhanced tracking protections. We may need to update this error message asking the user to try turning off ad blockers and tracking protection. 

It's a little awkward though.

radishmouse:
After discussing with @MelissaAutumn and @inattee, we're dropping this approach in favor of customizing the MC page

PR #1104, opened 2026-01-21
Add distinct title & description to each tbpro page

### LINKED ISSUE #1094: All tbpro pages use the same title/description meta
It's only set to _"Thunderbird Pro: Email Hosting, Scheduling & Secure Sharing"_ (title unlocalized) & _"Elevate your Thunderbird email experience with Thunderbird Pro's cloud based tools: secure email hosting via Thundermail, effortless meeting scheduling with Appointment, and encrypted file sharing through Send. Support free, open-source development."_ (description extracted for l10n) on the base include:

https://github.com/thunderbird/thunderbird-website/blob/f5f8d9126851f396c9eb4ee2fb5eac5816f6a70e/sites/tb.pro/includes/base/base.html#L10-L11

and then never overridden later on individual page — making their titles on the individual pages all be the same:

> Image

unable to distinguish them in tabs, and also all ending up with the same content in social sharing previews.

janbrasna:
Adds custom title & description metadata for every tbpro page, and wraps strings for localization, incl. little tweaks of current formatting.

The content is taken from existing strings (header for title and intro paragraph for description), save for some waitlist simplification, and some customizing around the inherited block defaults vs. thundermail page to add some extra keywords in there while still differentiating from the home page.

This also rearranges the title jinja blocks to be able to override the title fully, incl. the suffix, to avoid unwanted repetition if need be.

Resolves #1094

### DIFF
--- a/sites/tb.pro/waitlist/index.html
+++ b/sites/tb.pro/waitlist/index.html
@@ -5,7 +5,8 @@
 {% extends "includes/base/page.html" %}
 {% from 'includes/macros/right-arrow-submit-button.html' import right_arrow_submit_button %}
 
-{% block page %}{{ _('Join the Waitlist') }}{% endblock %}
+{% block page_title %}{{ _('Join the Waitlist') }}{% endblock %}
+{% block page_desc %}{{ _('Join the waitlist to get early access to Thunderbird Pro Early Bird.') }}{% endblock %}
 
 {% block extra_css %}
 <link href="{{ static('css/waitlist.css') }}" rel="stylesheet" type="text/css" />


janbrasna:
Correct. I expected given no i18n is enabled on the tb.pro site the strings are not exported yet, or not set up in Pontoon (in a new project? or within WTN project?) — I'll go check the l10n repos and report back if that went through localizers already.

MelissaAutumn:
Huh, I thought it was enabled, guess not. (I've been away from tb.net work :sweat_smile:) So no worries there then!

janbrasna:
Maybe I'm assuming wrong, but I don't see a "tb.pro-l10n" repo or Pontoon project of sorts, so expecting all of the strings to fall under "thunderbird.net-l10n" and its existing Pontoon project I'm pretty positive there's no tb.pro content exported and exposed to localizers at this point:

https://github.com/thunderbird/thunderbird.net-l10n/blob/df09b84f/fy_NL/LC_MESSAGES/messages.po

(e.g. a search for "tb.pro" in the repo returns empty; I'd expect to appear mentioned in the filenames referencing the strings — so it's apparently not wired yet anywhere…)

inattee:
I recall something was discussed about not using volunteer localizers for TB Pro since it's a paid product. I think we're going to do localization through Smartling. @Sancus would probably know more.  

Title/description text looks good. Thanks!

PR #1106, opened 2026-01-23
Canonical and alternate links updated

yavuzyazici:
When we check the website

All the URLs forwards to the version which has '/' at the end.

https://www.thunderbird.net/tr/donate => https://www.thunderbird.net/tr/donate/

So canonical tags and alternate URL addresses needs to be same with that. Some search engines are smart enought to understand that link difference. But some can't.

I thought to fix that from builder.py but i am not experienced enaught for this project. Maybe there is some consequences for canonical_path change

image


Sancus:
Hi, thanks for reporting the issue and for the pull request.

This isn't quite right though as it will produce extra `/` on some pages. Better changes would be in `builder.py` like so

Line 490 - change from:
`canonical_path = f'/{dir_part}' if dir_part else '/'`
to:
`canonical_path = f'/{dir_part}/' if dir_part else '/'`

Lines 311 and 318 - add trailing slash:
`self._env.globals['canonical_path'] = '/' + str(pathlib.Path(...)) + '/'`

However we should also add some tests to make this stays fixed and also make sure it doesn't mess anything up.

yavuzyazici:
Sorry for late reply. I was having busy days at office

Done.

PR #1129, opened 2026-02-16
fix padding on subscription plan (TB Pro landing page)

### LINKED ISSUE #1121: TB Pro landing page: reduce padding on subscription plan
Large breakpoint

Reduce the padding on the inner box of subscription plan (Early Bird Plan). Should be 24px top and bottom, 16px left and right - so that text has more room.

Image

aho15:
Resolves #1121 
Adjusts padding according to the issue.

### DIFF
--- a/assets/less/tbpro/index.less
+++ b/assets/less/tbpro/index.less
@@ -699,7 +699,7 @@ section#details {
             border-radius: @space-12;
             margin-block: @gutter;
             margin-inline: @space-36;
-            padding: @space-36;
+            padding: @space-24 @space-16;
             text-align: center;
             width: 100%;
 


assets/less/tbpro/index.less
janbrasna:
```suggestion
            padding: @space-36;
            text-align: center;
            width: 100%;

            @media (min-width: @lg) {
                grid-column: 1 / 4;
                padding: @space-24 @space-16;
            }
```

janbrasna: The ask might be something more along the lines:

```suggestion
            padding: @space-36;
            text-align: center;
            width: 100%;

            @media (min-width: @lg) {
                grid-column: 1 / 4;
                padding: @space-24 @space-16;
            }
```

janbrasna (commented):
The ticket sounds like _only_ changing the values for LG breakpoint, so might be worth confirming.

Issue #884, opened 2025-07-23
Holiday Calendar inconsistency: noun or adjective


hex-m:
Most calendars on the [list](https://www.thunderbird.net/en-US/calendar/holidays/) use the noun for the country (e.g. Australia), but some use the adjective (e.g. Austrian).

It seems the noun is the more common case so I assume it should be standardized to use the noun.

Cases that don't use the noun:
* Austrian
* Belgian
* Czech
* Polish
* Swedish

I noticed because Nextcloud shows the calendar in the UI as

> Holidays in Austrian

MelissaAutumn:
Thanks! I was mostly following the old convention, added some tracking labels to this ticket.

PR #1130, opened 2026-02-19
Changed some calendar names to represent country names

aho15:
Corresponding issue: #884
Changed calendar names to the respective country name instead of language in settings.py.

Unfortunately, due to free tier limitation, I am not able to update all calendars.
This should be done as they currently contain holidays from 2025 - 2028 instead of from 2026 - 2029 and they seem to have been last updated in 04/2025.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -608,40 +608,26 @@ CALENDAR_LOCALES = {
 
 # Used to normalize filenames for calendar generation
 # This is the first part of the filename, 'Holidays' is bolted on in code.
-# So 'Belgium (Dutch)' will become 'BelgiumHoldays.ics'
+# So 'Belgium (Dutch)' will become 'BelgiumHoldaysDutch.ics'
 # If the value is a tuple, the first part is before 'Holidays' and the second part is after.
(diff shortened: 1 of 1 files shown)


settings.py
janbrasna: The question is whether to also not stop turning the country names into the adjectives here in the "normalization" phase… ("Italian", "French", "German" etc. above too — as not just the calendar names earlier, but also this mapping step changes the end result too…)

> filename

janbrasna (commented):
This has the potential to result in broken links for consumers — who would have to figure out something's wrong with their ics subscription, come back to the site, and subscribe to the correct file now… so I figure the actual tricky part of the ticket for any renaming really is the necessary addition of all the old-to-new redirects in some place as .htaccess (or its build script)…

aho15:
I am sorry for costing your time. I should have seen both of these issues and at least bring them up to you.

Concerning the mapping of the filenames: I removed the entries which only led to change from country name to language. I then recognized for Algeria, it's default language was french, which made me think: why not including the default languages names in the file name as well? I think it won't hurt further and increases consistency.

PR #1137, opened 2026-02-25
Serve new autoconfig JSON file on the root path

babolivier:
As of https://github.com/thunderbird/autoconfig/pull/180, the autoconfig deployment script generates a JSON file listing all generated configurations (`generated_files.json`). That script (`tools/convert.py` in the autoconfig repo) doesn't have knowledge of the folder hierarchy on the server once deployed, and so generates everything in a single folder.

After chatting with @Sancus, we agreed that it would be better to serve this file outside of the `v1.1/` path so we don't risk confusing existing tools that consume ISPDB, which might expect everything under `v1.1/` to be XML.


Sancus:
Thanks!

PR #1143, opened 2026-03-05
Add generic redirect for url formatter support links. (Fixes #1135)

### LINKED ISSUE #1135: Add redirects for URLs generated by the url formatter
In some locations we use

```
Services.urlFormatter.formatURLPref("app.support.baseURL") + slug
```

or the `moz-support-link` custom element, which both use the following URL pattern:

```
https://support.thunderbird.net/thunderbird/<version>/<os>/<language>/<slug>
```

which are currently not properly redirected. They should redirect to

```
https://support.mozilla.org/<language>/kb/<slug>
```

Examples:
1. [broken] https://support.thunderbird.net/thunderbird/147.0.2/Linux/en-US/quarantined-domains -> https://support.mozilla.org/en-US/kb/quarantined-domains
2.  [broken] https://support.thunderbird.net/thunderbird/147.0.2/Linux/en-US/experiment-add-on-support -> https://support.mozilla.org/en-US/kb/experiment-add-on-support
3. [custom] https://support.thunderbird.net/thunderbird/147.0.2/Linux/en-US/extension-permissions -> https://support.mozilla.org/en-US/kb/permission-request-messages-thunderbird-extensions

The last example marked as [custom] is just FYI. It is a manually added redirect, where we do not redirect to the generic KB URL which points to a Firefox branded page, but to our own version with a different slug.

Example 1 will point to a Firefox branded page, Example 2 will point to a Thunderbird branded page. So the URL itself will not determine if the destination is a Firefox or Thunderbird page, that is defined for each slug individually on SUMO. For URLs in toolkit code (which will most probably point to Firefox branded pages), we have to determine individually, if we are ok to point to the Firefox branded page or not (as shown in Example 1 & 3). For pages created specifically for us, we definitely need the generic redirect, as shown in Example 2.

Sancus:


jobisoft:
You did include the language in the redirect. I am also fine with not doing that as we did it with all the other `/kb/` redirects. Whatever feels better for you.

Thanks for the fix!

jobisoft (approved):

PR #1147, opened 2026-03-09
Use uv for dependency management and clean up requirements and docs. (Fixes #864, Fixes #1146)

### LINKED ISSUE #864: Update readme's build instructions
We should include instructions on making a virtual environment with at least python 3.12.

### LINKED ISSUE #1146: move to uv
we should use uv as project/package manager

Sancus:


pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+    "requests~=2.32.4",
+    "setuptools==78.1.1",
+    "watchdog>=0.9.0",
+    "webassets>=0.12.1",
+    "webob>=1.7.3",
+    "pytest~=7.3.1",
+]
+
+[dependency-groups]
+dev = ["flake8>=3.7.7"]
```
MelissaAutumn, line 27: I think we'll want to move to ruff like all of other projects. But this is fine for now.

pyproject.toml
```diff
@@ -0,0 +1,34 @@
+[project]
+name = "thunderbird-website"
+version = "0.1.0"
```
MelissaAutumn, line 3:
```suggestion
version = "1.0.0"
```

PR #1159, opened 2026-03-12
Roadmaps: Ensures that bolt-button uses dark-mode color

radishmouse:
Makes sure bolt buttons use correct color value. 
(Sets CSS variable value to dark-mode color.)


radishmouse:
One line change, specifying a color value. Skipping code review for deployment to stage.

PR #1179, opened 2026-03-24
Create a preview URL for each non-fork pull request

Sancus:
### PR Preview Deployments
Adds per-PR preview environments using AWS Lambda + API Gateway behind `{branch-name}-{pr number}.thunderbird.dev`. For example `previews-1179.thunderbird.dev`

### Usage
Add a label to trigger a preview:

`preview` — builds www.thunderbird.net
`preview:startpage`, `preview:updates`, `preview:tbpro`, `preview:roadmaps` — builds the corresponding site
Subsequent pushes auto-rebuild while the label is present. Closing the PR tears down all resources.

### DIFF
--- a/tests/test_webserver.py
+++ b/tests/test_webserver.py
@@ -148,6 +148,7 @@ def test_locale_detection(host: str, path: str, lang: str, expected_location: st
 MIME_TYPE_TESTS = [
     ("/media/img/thunderbird/base/about/privacy.avif", "image/avif"),
     ("/media/img/thunderbird/base/about/privacy.webp", "image/webp"),
+    ("/media/img/thunderbird/base/about/privacy.jxl", "image/jxl"),
 ]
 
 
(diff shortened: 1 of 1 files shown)


.github/workflows/preview.yml
```diff
@@ -0,0 +1,197 @@
...
+                issue_number: context.issue.number,
+                body,
+              });
+            }
+
+  destroy-preview:
+    if: >-
+      github.event.action == 'closed' &&
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      contains(join(github.event.pull_request.labels.*.name, ','), 'preview')
```
MelissaAutumn: I think if we remove the preview label and merge this we'll leave a preview environment stuck up forever (or until manually brought down.)

I think this should be `if closed OR preview contained preview and no longer does`. Within the check you should check if the lambda exists, and if it does then tear it down. (If destroy preview infra does that then rad.)

Sancus: Hmm. This workflow triggers on unlabeled,  and so if you remove the label it should destroy the preview environment. I think.

Sancus: Well that's how it was supposed to work, but it was not working. I think I've fixed it now.

PR #1187, opened 2026-03-31
[Appeal April 2026] Make sun ray height be full screen

davinotdavid:


davinotdavid:
Sneaking this in as it was confirmed by @inattee and @laurelterlesky that the sun rays are fixed in the preview link!

PR #1200, opened 2026-04-25
Fix hover underline for footer legal links (#1132)

### LINKED ISSUE #1132: TB Pro Landing Pages - Footer - hover states
Issue: there aren't any hover states in the footer

Intended behaviour:
Hover states should have an underline

micheal000010000-hub:
### Description

Fixes missing hover highlight for footer links in the legal section.

### Changes

* Added hover/focus styles for `#tb-legal a`
* Updated underline color on hover to provide visible feedback
* Scoped changes to legal section to avoid affecting other footer links

### Testing

* Ran local build using `uv run build-site.py --watch`
* Verified underline highlight on hover for:

  * MZLA Technologies Corporation
  * Creative Commons license
  * Contribute to this site
* Confirmed existing footer links remain unchanged

### Fixes

Closes #1132

### DIFF
--- a/assets/less/common/common-footer.less
+++ b/assets/less/common/common-footer.less
@@ -111,7 +111,7 @@ footer#tb-footer {
     }
 }
 
-#tb-legal {
+footer#tb-footer #tb-legal {
     font-size: .75rem;
 
     a {
--- a/assets/less/common/common-footer.less
+++ b/assets/less/common/common-footer.less
@@ -115,6 +115,7 @@ footer#tb-footer #tb-legal {
     font-size: .75rem;
 
     a {
+        --accent: #1E4EAE;
         text-decoration: underline;
         text-decoration-color: currentColor;
 
--- a/assets/less/common/common-footer.less
+++ b/assets/less/common/common-footer.less
@@ -113,7 +113,14 @@ footer#tb-footer {
 
 #tb-legal {
     font-size: .75rem;
+
     a {
         text-decoration: underline;
+        text-decoration-color: currentColor;
+
+        &:hover,
+        &:focus {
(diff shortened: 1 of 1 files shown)


assets/less/common/common-footer.less
```diff
@@ -113,7 +113,14 @@ footer#tb-footer {
 
 #tb-legal {
```
radishmouse: Thanks for the fix! I think you need to add this to make sure that the `text-decoration: none` from `footer#tb-footer a` doesn't clobber the new text decoration:

```suggestion
footer#tb-footer #tb-legal {
```

assets/less/common/common-footer.less
```diff
@@ -113,7 +113,14 @@ footer#tb-footer {
...
     font-size: .75rem;
+
     a {
         text-decoration: underline;
```
radishmouse, line 119: Also need a definition for the `--accent`:
```suggestion
        --accent: #1E4EAE;
        text-decoration: underline;
```

radishmouse (changes requested):
Thanks for the changes! I've suggested a couple of tweaks to get it working against the latest version of the site code.

assets/less/common/common-footer.less
```diff
@@ -111,9 +111,17 @@ footer#tb-footer {
     }
 }
 
-#tb-legal {
+footer#tb-footer #tb-legal {
```
radishmouse, line 114: Ideally, we would move the `#tb-legal` styles into `footer#tb-footer`, but I left them here so that the suggestions were in context of your original changes.

radishmouse:
I've gone ahead and incorporated the suggestions. Will merge on your behalf. Thank you for the fix!

PR #1206, opened 2026-05-04
Route well-known paths at tb.pro domain to mail servers

ryanjjung:
Adds a rewrite rule to the httpd config which will send requests originally bound for `tb.pro/.well-known/whatever` to `mail.thundermail.com/.well-known/whatever` instead.

### DIFF
--- a/docker/conf/ssl.conf
+++ b/docker/conf/ssl.conf
@@ -336,6 +336,20 @@ ScriptSock /tmp/cgisock
   Header always set X-Frame-Options "DENY"
 </VirtualHost>
 
+# thundermail.com
+<VirtualHost *:8080>
+  ServerName https://thundermail.com
+  ServerAlias www.thundermail.com stage-thundermail.com
+
+  # Rewrite well-known paths to our mail server
+  # https://github.com/thunderbird/mailstrom/issues/213
+  RewriteEngine On
+  RewriteRule ^/.well-known/(.*)$ https://mail.thundermail.com/.well-known/$1
+
+  # Rewrite all other requests to tb.pro site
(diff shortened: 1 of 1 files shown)


ryanjjung:
Don't think this is a requirement. Moving to draft, might close later.

ryanjjung:
The tb.pro stuff is not explicitly required, but we do need to accomplish redirects from the base `thundermail.com` domain (as opposed to the `mail.thundermail.com` subdomain which actually serves these files). DNS for thundermail.com routes to thunderbird.net, so I added a rewrite rule for that domain's well-known addresses, pointing them to `mail.thundermail.com`. The tb.pro rule, though not necessary, doesn't hurt, so I've left it. Happy to remove it if someone else thinks that's a good idea.

docker/conf/ssl.conf
```diff
@@ -337,6 +342,11 @@ ScriptSock /tmp/cgisock
   ServerAlias www.tb.pro www-stage.tb.pro stage.tb.pro
   DocumentRoot /var/www/html/start/tb.pro
 
+  # Rewrite well-known paths to our mail server
```
Sancus, line 354: I would add a proper virtualhost for www.thundermail.com similar to the one for tb.pro

Sancus (changes requested):

ryanjjung:
Alright, I added a new vhost for thundermail.com. We don't have any specific content to serve for that domain, so I added the two sets of redirects:

- Requests for /.well-known/anything will go to the mail host instead.
- Requests for anything else will route striaght to https://tb.pro/

PR #1208, opened 2026-05-04
Let SUMO decide the language for the /kb/ask redirect

freaktechnik:
I don't think we should hard code the language code.


Sancus:
I think the reason we did this might have been that you would likely get no answer if you submitted a question in most non-English languages.

But I agree it no longer makes sense.

PR #1260, opened 2026-07-09
Don't do locale redirects for tb.pro for now

davinotdavid:
We don't have translation strings for tb.pro just yet and it is still a topic of discussion on how to achieve this so this PR forces the locale to be en-US instead so that the pages don't look half-translated when we use shared strings.


davinotdavid:
@Sancus coming from the website meeting earlier today, this was a carried over item that I wasn't sure if it was me or you that should tackle this. Is this what you intended?

PR #1267, opened 2026-07-20
Pause animations in appeals if window / tab not in focus

### LINKED ISSUE #1266: Pause animation appeal if the tab is not focused
We're getting some reports of high cpu usage on our appeal pages. While removing the animation is not possible in cases where the page was designed around it, we could at least use javascript ([focus](https://developer.mozilla.org/en-US/docs/Web/API/Window/focus_event) and [blur](https://developer.mozilla.org/en-US/docs/Web/API/Window/blur_event)) to detect if the tab is in focus/blur. This could then append or remove a page level css class.

With that css class as the main selector, followed by the page we could probably use [animation-play-state](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation-play-state) to pause the animation. I would have thought browsers did this automatically but I guess not. 

Bugzilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=2038287
https://bugzilla.mozilla.org/show_bug.cgi?id=2007785

davinotdavid:
## Description of changes

- Added an `animation-visibility.js` script to add an event handler and toggle the `tab-inactive` class in the `documentElement`.
- Updated donation button CSS, `128.0/apr25/` and `140.0/apr26-1a/` and `140.0/apr26-1c/

Relevant preview links:
https://davinotdavid-pause-animations--1267-whimbrel.thunderbird.dev/en-US/thunderbird/128.0/apr25/
https://davinotdavid-pause-animations--1267-whimbrel.thunderbird.dev/en-US/thunderbird/140.0/apr26-1a/
https://davinotdavid-pause-animations--1267-whimbrel.thunderbird.dev/en-US/thunderbird/140.0/apr26-1c/

## Limitations and notes

- This does not improve the performance while the tab is active. There's a much longer / deeper research and refactor required for this that's out of scope of this task.

## Videos

Focus on the tab vs focus on DevTools toggles the pause animation:

https://github.com/user-attachments/assets/1e4f2c18-0e77-45a8-80c9-a8f77acc41df

Profiling GPU process in Chrome by switching tabs:

https://github.com/user-attachments/assets/b632901d-367b-45fe-87b1-06e5fed608c4

## Related issues

Closes https://github.com/thunderbird/thunderbird-website/issues/1266

### DIFF
--- a/assets/less/appeals/apr25.less
+++ b/assets/less/appeals/apr25.less
@@ -56,6 +56,11 @@
       background: var(--bg-img);
       background-size: 100% 100%;
     }
+
+    // Pause the gradient animation while the tab isn't visible or focused.
+    :root.tab-inactive & {
+      animation-play-state: paused;
+    }
   }
 
   #appeal-header {
--- a/assets/less/appeals/apr25.less
+++ b/assets/less/appeals/apr25.less
@@ -40,6 +40,11 @@
 // Scope to just the appeal page
 .page-appeal-apr25 {
 
+  // Pause the gradient animation while the tab isn't visible or focused.
+  &.tab-inactive body {
+    animation-play-state: paused;
+  }
+
   body {
     margin: auto;
     min-width: var(--min-small-screen);
@@ -55,11 +60,6 @@
       animation: none;
       background: var(--bg-img);
       background-size: 100% 100%;
-    }
-
-    // Pause the gradient animation while the tab isn't visible or focused.
-    :root.tab-inactive & {
-      animation-play-state: paused;
     }
   }
 
--- a/assets/less/appeals/apr26-1e.less
+++ b/assets/less/appeals/apr26-1e.less
@@ -8,6 +8,11 @@
 .page-appeal-dec25 {
   background-color: #ffc068;
 
+  // Pause the gradient animation while the tab isn't visible or focused.
+  &.tab-inactive body {
+    animation-play-state: paused;
+  }
+
   body {
     max-width: none;
     min-width: var(--min-small-screen);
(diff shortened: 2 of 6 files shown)


MelissaAutumn:
Seems like the gradient backdrop on https://davinotdavid-pause-animations--1267-whimbrel.thunderbird.dev/en-US/thunderbird/128.0/apr25/ has some issues where sometimes it pauses but most of the time it doesn't. 

On chrome via leaving that tab active in the background but switching OS focus to a new window. :thinking: I believe the tab-inactive code is working correctly here. And switching to a different chrome tab does pause it correctly (as noticed by cpu activity.)

Edit: The chrome tab switching might just be chrome pausing animations. I think the css query needs a tweak.

assets/less/appeals/apr25.less
```diff
@@ -56,6 +56,11 @@
       background: var(--bg-img);
       background-size: 100% 100%;
     }
+
+    // Pause the gradient animation while the tab isn't visible or focused.
+    :root.tab-inactive & {
```
Sancus: apr25 background animation does not actually pause with this selector

Needs to be moved outside the nested body and
```
&.tab-inactive body {
  animation-play-state: paused;
}
```

assets/less/appeals/apr26-1e.less
```diff
@@ -20,6 +20,11 @@
       background: var(--bg-img);
       background-size: 100% 100%;
     }
+
+    // Pause the gradient animation while the tab isn't visible or focused.
+    :root.tab-inactive & {
```
Sancus: Same problem as apr25.

Sancus (changes requested):
We should probably have a way to test stuff like this but I'm not blocking this quick PR on that.

Approved after those fixes.

MelissaAutumn (approved):
Works for me now, thanks!

davinotdavid:
@Sancus @MelissaAutumn Whoops! Thank you for the reviews, I guess I was too focused on the donation button 😅. 

Agreed on having a better way to test this so cut a ticket here just so we don't forget about it: https://github.com/thunderbird/thunderbird-website/issues/1268

PR #1282, opened 2026-08-06
Update l10n tools

radishmouse:
This PR updates the tooling for localization:

When running the updated `extract.sh`:
- Stops immediately if extraction, merging, or formatting fails.
- Runs Python tooling through the project’s uv environment.
- Normalizes formatting across every locale catalog to reduce noisy, formatting-only diffs.

The PR also includes a read-only checker that compares generated PO files with their Git baseline. It distinguishes definite translation loss from expected source changes and new untranslated strings, without modifying translations.

## Testing
- Added 11 focused checker tests.
- Full test suite passes.
- Verified against the current Roadmaps extraction across 99 locale catalogs.


Sancus:
I don't really like adding AGENTS.md to this repository purely as instructions about l10n. Really, extraction is something that should happen automatically in CI when new strings are added. There is also a lot of complexity in these instructions, I'll review them in more detail later but on first glance I don't know how there is so much for what is a very simple process.

radishmouse:
> I don't really like adding AGENTS.md to this repository purely as instructions about l10n. Really, extraction is something that should happen automatically in CI when new strings are added. There is also a lot of complexity in these instructions, I'll review them in more detail later but on first glance I don't know how there is so much for what is a very simple process.

All good - I'm removing the AGENTS.md, particularly since it includes steps for building the roadmaps site.

It also included instructions for doing the following:
- making sure `libs/locale` is up to date
- creating a backup branch of `libs/locale`
- run the extraction and saving a log
- run the clobber checking
- check the diffs for a few of the main translations (DE, FR, JA, etc.)
- running the tooling tests
- reporting on what happened

PR #1290, opened 2026-08-26
Update package dependencies to fix security vulnerabilities - August 26th 2026

davinotdavid:
Updated the following:

pyproject.toml — bumped pinned/floor versions:
  - pymdown-extensions ==10.21 → ==11.0.1
  - setuptools ==82.0.1 → ==83.0.0
  - webob >=1.7.3 → >=1.8.10
  - pillow ==12.2.0 → ==12.3.0

uv.lock (root):
  - idna 3.11 → 3.19
  - urllib3 2.6.3 → 2.7.0
  - soupsieve 2.8.3 → 2.9.2

## QA Log
- Running `uv audit` runs clean, no vulnerabilities
- Tests pass
- Local build works as usual
- Image tools that use Pillow works as usual


radishmouse (approved):

radishmouse:
Note: the website is static HTML, and these deps only affect local development machines (where we generate the static HTML). 

I can't imagine an attack vector other than general malware or blanket data exfiltration. However, it's probably a good idea for us to keep our deps relatively fresh so that we don't break the build system at some point in the future.

Sancus:
> Note: the website is static HTML, and these deps only affect local development machines (where we generate the static HTML).
> 
> I can't imagine an attack vector other than general malware or blanket data exfiltration. However, it's probably a good idea for us to keep our deps relatively fresh so that we don't break the build system at some point in the future.

CI runs the builds as well so there are certain types of vulns that could have an impact there. But yes, you're totally correct that 99.99% of the time the vulns don't really matter to this project.

PR #1311, opened 2026-09-16
Make legal downloader handle policies from Thundermail

kewisch:
I didn't want to change these by hand, so I adjusted the legal downloader.


builder.py
```diff
@@ -103,24 +104,52 @@ def delete_contents(dirpath):
...
-
-        contents = requests.get(settings.THUNDERBIRD_DESKTOP_PRIVACY_POLICY_URL).text
-        html = markupsafe.Markup(markdown.markdown(contents, extensions=['markdown.extensions.attr_list']))
+        for destination, source in self.DOCUMENTS.items():
+            logger.info('Reading %s to %s', source, destination)
 
-        html = f'{{# THIS PAGE IS AUTOMATICALLY GENERATED, DO NOT EDIT THIS DOCUMENT! #}}\n{html}'
+            contents = self.read_document(source)
+            html = markupsafe.Markup(
+                markdown.markdown(contents, extensions=['markdown.extensions.attr_list'])
```
Sancus: The tb.pro privacy policy contains a pipe table (cookie types), but only `attr_list` is enabled, so it renders as a literal paragraph in `sites/tb.pro/includes/legal/privacy.html`. The hand-written version this replaces had a real `<table>`.

Add `'markdown.extensions.tables'` here and regenerate.

Cookie table rendered as literal markdown

sites/tb.pro/includes/legal/privacy.html
```diff
@@ -0,0 +1,179 @@
+{# THIS PAGE IS AUTOMATICALLY GENERATED, DO NOT EDIT THIS DOCUMENT! #}
+<h1>MZLA Privacy Statement</h1>
```
Sancus: This gives both tb.pro pages two `<h1>`s: the template header still has `<h1>{{ _('Privacy Policy') }}</h1>` / `<h1>{{ _('Terms') }}</h1>`, and the include adds another. Either drop the header-block `<h1>` or demote the generated one to `<h2 id="page-header">` like before (which would also make the `plain.less` change unnecessary).

Sancus (changes requested):
Approach looks good, two things to fix before merge.

sites/tb.pro/includes/legal/privacy.html
kewisch: I added some code to demote the headings. It is kinda weird, the h1 already on the page is actually invisible through CSS, so in theory now the h2 is the top visible heading.

